PR #13, opened 2021-12-16
Update guide URL and remove status

gsmet:


gsmet:
@middagj I will release a new version before I ship 2.6.0.Final as I want the guide thing out there.

PR #16, opened 2021-12-21
Remove maven-settings.xml.gpg

gastaldi:
This removes the encrypted file from the repository and uses the recommended approach in GitHub to publish to Maven Central
See https://docs.github.com/en/actions/publishing-packages/publishing-java-packages-with-maven#publishing-packages-to-the-maven-central-repository


middagj:
Can we merge this one?

zhfeng:
@middagj I think we are safe to merge this PR.

PR #35, opened 2022-04-01
Update to Artemis 2.21.0 and remove redundant exclusions

middagj:


gsmet:
You positive that Johnzon is not present because things get very messy when it’s in the classpath.

middagj:
Yes, I compared the dependency:tree output. They now use a shaded version, see https://github.com/apache/activemq-artemis/commit/bfc10dcf8bcb1663519fb74907623a5f6b1d0710

middagj:
If I remember correctly, I added those after failing native integration test. And they succeed now.

middagj:
I thought about this and considered the possibility users might want to use the newer extension but with an older artemis-client due to compatibility reaons (although I think they make sure new clients can talk with older servers).

PR #45, opened 2022-06-10
Fix #44 to support XA Connection Factory

### LINKED ISSUE #44: Support XAConnectionFactory
I'd like introduce an option `quarkus.artemis.xa=[true|false]` and if it is `true`, `ArtemisJmsProcessor` will produce a `XAConnectionFatory` bean.

```
public Supplier<XAConnectionFactory> getXAConnectionFactorySupplier() {
        return new Supplier<XAConnectionFactory>() {
            @Override
            public XAConnectionFactory get() {
                return new ActiveMQXAConnectionFactory(config.url, config.username.orElse(null), config.password.orElse(null));
            }
        };
    }
```

zhfeng:


integration-tests/jms/pom.xml
```diff
@@ -18,6 +18,12 @@
             <artifactId>quarkus-resteasy</artifactId>
         </dependency>
 
+        <!-- Narayana Transaction Manager -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-narayana-jta</artifactId>
+        </dependency>
```
middagj, line 25: I think this should be optional

integration-tests/jms/src/main/java/io/quarkus/it/artemis/ArtemisProducerManager.java
```diff
@@ -5,17 +5,51 @@
...
     public void send(String body) {
         try (JMSContext context = connectionFactory.createContext(JMSContext.AUTO_ACKNOWLEDGE)) {
-            JMSProducer producer = context.createProducer();
-            producer.send(context.createQueue("test-jms"), body);
+            _send(context, body);
+        }
+    }
+
+    public void sendXA(String body) throws Exception {
+        XAJMSContext context = xaConnectionFactory.createXAContext();
```
middagj, line 37: shouldn't this also be in a try with resources?

middagj (commented):
Some small things, but looks good to me.

integration-tests/jms/pom.xml
zhfeng: Good catch!

integration-tests/jms/src/main/java/io/quarkus/it/artemis/ArtemisProducerManager.java
zhfeng: Anyway, this should be better processed by other wrapper, such as `org.messaginghub:pooled-jms`.

zhfeng:
@middagj I also need it for working on https://github.com/quarkusio/quarkus/issues/14871

integration-tests/jms/src/main/java/io/quarkus/it/artemis/ArtemisProducerManager.java
middagj: Okay, not an expert on XA and it is test code, so should be fine.

PR #55, opened 2022-06-26
Fix #54 to add DevServices for Artemis JMS

### LINKED ISSUE #54: Add DevService support
It could be helpful to start an ActiveMQ Artemis broker in dev mode automatically.
Although we have `ArtemisTestResource`, it's only used in the test mode.

zhfeng:


zhfeng:
Hi @middagj ,

Can you review these changes? and I'd like to include them in a new release if it makes sense.

Thanks!

middagj:
Sure, it can take a few days

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisDevServicesBuildTimeConfig.java
```diff
@@ -0,0 +1,71 @@
...
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class ArtemisDevServicesBuildTimeConfig {
+
+    /**
+     * If Dev Services for ActiveMQ Artemis has been explicitly enabled or disabled. Dev Services are generally enabled
+     * by default, unless there is an existing configuration present. For Artemis, Dev Services starts a broker unless
```
middagj, line 13: I had to read this sentence a few times before I got the idea
> Dev Services are generally enabled by default, unless there is an existing configuration present.
This is explained in the following sentence? As in, if you are upgrading from a previous version, it won't run. I would make this sentence more future proof.

jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/DevServicesArtemisJmsProcessor.java
```diff
@@ -0,0 +1,231 @@
...
+import io.quarkus.deployment.logging.LoggingSetupBuildItem;
+import io.quarkus.devservices.common.ConfigureUtil;
+import io.quarkus.devservices.common.ContainerAddress;
+import io.quarkus.devservices.common.ContainerLocator;
+import io.quarkus.runtime.configuration.ConfigUtils;
+
+/**
+ * Start a ActiveMQ Artemis broker if needed
+ */
+public class DevServicesArtemisJmsProcessor {
```
middagj, line 35: Can we get this in core and create something that is applicable for both core and JMS?

jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/DevServicesArtemisJmsProcessor.java
```diff
@@ -0,0 +1,231 @@
...
+                    !devServicesSharedNetworkBuildItem.isEmpty(),
+                    devServicesConfig.timeout);
+            if (devService == null) {
+                compressor.closeAndDumpCaptured();
+            } else {
+                compressor.close();
+            }
+        } catch (Throwable t) {
+            compressor.closeAndDumpCaptured();
+            throw new RuntimeException(t);
```
middagj, line 87:
```java
throw (t instanceof RuntimeException ? (RuntimeException) t : t
```

middagj (commented):
I like the addition, some comments.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisDevServicesBuildTimeConfig.java
zhfeng: Well, I borrow the sentence from the `KafkaDevServicesBuildTimeConfig` and I think it means that

>The dev servies for ActiveMQ Artemis is enabled by default unless `quarkus.artemis.url` is set or `quarkus.artemis.devservices.enabled` is set to `false` explictly.

jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/DevServicesArtemisJmsProcessor.java
zhfeng: I think it should be wrapped with `RuntimeException`.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisBuildTimeConfig.java
```diff
@@ -18,4 +18,10 @@ public class ArtemisBuildTimeConfig {
      */
     @ConfigItem(name = "xa.enabled", defaultValue = "false")
     public boolean xaEnabled;
+
+    /**
+     * Configuration for DevServices. DevServices allows Quarkus to automatically start ActiveMQ Artemis in dev and test mode.
+     */
+    @ConfigItem
+    public ArtemisDevServicesBuildTimeConfig devservices;
```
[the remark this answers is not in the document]
zhfeng: Hmm, sorry, I think we have to leave it as `devservices` to be consistent with the global devservices config name.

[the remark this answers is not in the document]
middagj: Ah, better be consistent then.

jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/DevServicesArtemisJmsProcessor.java
middagj: Ah yes, I missed that. Should be:

```java
throw (t instanceof RuntimeException ? (RuntimeException) t : new RuntimeException(t);
```

the idea is to not unnecessarily wrap exceptions in `RuntimeException`.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -0,0 +1,229 @@
...
+                container.withLabel(DevServicesArtemisProcessor.DEV_SERVICE_LABEL, config.serviceName);
+            }
+
+            timeout.ifPresent(container::withStartupTimeout);
+
+            container.start();
+            return new RunningDevService("ActiveMQ-Artemis",
+                    container.getContainerId(),
+                    container::close,
+                    QUARKUS_ARTEMIS_URL, String.format("tcp://localhost:%d", container.getMappedPort(ARTEMIS_PORT)));
```
middagj, line 174:
```java
String.format("tcp://%s:%d", container.getHost(), container.getMappedPort(ARTEMIS_PORT))
```

integration-tests/jms/src/main/resources/application.properties
```diff
@@ -1,2 +1,2 @@
-quarkus.artemis.xa.enabled=true
-quarkus.artemis.url=tcp://localhost:61616
+quarkus.artemis.devservices.enabled=false
```
middagj, line 1: This should not be necessary, right?

middagj:
We should be able to also use testcontainers on core integration test. If you have time, could you add that, otherwise I will add it myself. Also we don't need the broker.xml files.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
zhfeng: Good catch!

integration-tests/jms/src/main/resources/application.properties
zhfeng: Well, it is compatible with using `ArtemisTestResouce` which starts/stops `EmbeddedActiveMQ`. I think we need to mention in the document that if users still want to use `ArtemisTestResource`, it needs to set `quarkus.artemis.devservices.enabled=false` or `quarkus.artemis.url` explictly.

IIRC we change to use `EmbeddedActiveMQ` dues to [vromero/activemq-artemis-docker](https://github.com/vromero/activemq-artemis-docker) does not maintain anymore.

I'm not sure that [artemiscloud/activemq-artemis-broker](https://quay.io/repository/artemiscloud/activemq-artemis-broker) could be supported by Red Hat offically? @gastaldi @gsmet 

I think we might add `@Deprecated` on `ArtemisTestResource` and will drop it in the future 2.x?

zhfeng:
yeah, I will add one in core integration test. And `broker.xml` is needed if we still use `ArtemisTestResource`.

middagj:
> yeah, I will add one in core integration test. And `broker.xml` is needed if we still use `ArtemisTestResource`.

correct, but only one

integration-tests/jms/src/main/resources/application.properties
middagj: [vromero was never active](https://github.com/vromero/activemq-artemis-helm/pull/25), we have been using our own images since then. Good point about `ArtemisTestResource`, it is actually used by some other projects. I have no clear view about the usage. We can keep both, it is a small class.

middagj:
Thanks for the PR

middagj:
Could you give it a test, and confirm it works as expected in your project. I will then release.

zhfeng:
Thanks @middagj and yeah I confirm it works in my camel-quarkus -example project. Please go heading for the release.

zhfeng:
@middagj I just realized that `ArtemisTestResource` is useful on windows where docker is not supported currently. So it should be better to remove `@Deprecated` . I'm sorry for the inconveniences.

middagj:
I didn't realize you added it, will remove it.

zhfeng:
I raise https://github.com/quarkiverse/quarkus-artemis/pull/60

PR #60, opened 2022-07-01
Remove @Deprecated on ArtemisTestResource

zhfeng:


test-framework/src/main/java/io/quarkus/artemis/test/ArtemisTestResource.java
```diff
@@ -10,7 +10,6 @@
 
 import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
 
-@Deprecated
```
middagj, line 13: You can remove the import as well.

PR #61, opened 2022-07-04
Improve Artemis DevService

zhfeng:
- remove `queues` property and add `extraArgs` which could be passed to container as `AMQ_EXTRA_ARGS`
- introduce `ArtemisContainer` to support fixedExposedPort

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisDevServicesBuildTimeConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisDevServicesBuildTimeConfig.java
@@ -1,7 +1,5 @@
 package io.quarkus.artemis.core.deployment;
 
-import java.util.Collections;
-import java.util.List;
 import java.util.Optional;
 
 import io.quarkus.runtime.annotations.ConfigGroup;
@@ -71,8 +69,8 @@ public class ArtemisDevServicesBuildTimeConfig {
     public String password;
 
     /**
-     * Queues to create on starting
+     * The value of the {@code AMQ_EXTRA_ARGS} environment variable to pass to the container.
      */
-    @ConfigItem(defaultValue = "[]")
-    public List<String> queues = Collections.emptyList();
+    @ConfigItem(defaultValue = "--no-autotune --mapped --no-fsync")
+    public String extraArgs;
 }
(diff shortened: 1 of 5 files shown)


middagj:
Why are we creating our own Artemis DevServices if there exists an AMQP DevServices which almost does the exact same thing except listening on a different port? Artemis can have multiple protocol on the same port.

zhfeng:
I think `AMQP DevServices` expose port `5672` which only supports `AMQP` protocol
```
2022-07-04 09:25:45,565 INFO  [org.apache.activemq.artemis.core.server] AMQ221020: Started EPOLL Acceptor at 172.17.0.6:61616 for protocols [CORE,MQTT,AMQP,STOMP,HORNETQ,OPENWIRE]
2022-07-04 09:25:45,575 INFO  [org.apache.activemq.artemis.core.server] AMQ221020: Started EPOLL Acceptor at 172.17.0.6:5445 for protocols [HORNETQ,STOMP]
2022-07-04 09:25:45,588 INFO  [org.apache.activemq.artemis.core.server] AMQ221020: Started EPOLL Acceptor at 172.17.0.6:5672 for protocols [AMQP]
2022-07-04 09:25:45,590 INFO  [org.apache.activemq.artemis.core.server] AMQ221020: Started EPOLL Acceptor at 172.17.0.6:1883 for protocols [MQTT]
2022-07-04 09:25:45,593 INFO  [org.apache.activemq.artemis.core.server] AMQ221020: Started EPOLL Acceptor at 172.17.0.6:61613 for protocols [STOMP]
2022-07-04 09:25:45,600 INFO  [org.apache.activemq.artemis.core.server] AMQ221007: Server is now live
2022-07-04 09:25:45,600 INFO  [org.apache.activemq.artemis.core.server] AMQ221001: Apache ActiveMQ Artemis Message Broker version 2.22.0 [broker, nodeID=68d07139-f8e6-11ec-b02d-0242ac110006] 

```
And if we want to use `AMQP DevServices`, we have to add the `smallrye-reactive-messaging-amqp` extension dependency. So we have to creat our own Artemis DevServices although most of the codes can be reused except the expose port.

zhfeng:
It should be better if quarkus can extract the classes in `AMQP DevServices` for reusing. I think there is a similar request with https://github.com/amqphub/quarkus-qpid-jms/issues/15

middagj:
Thanks for bringing it up there. AMQP could use port 61616 instead of default AMQP port 5672. It works the same.

PR #77, opened 2022-10-04
Release 1.3.0

### LINKED ISSUE #76: Release plan
Hello,

I would like to know whether there is a release plan for the Quarkus Artemis extension.
We possibly will need an updated Artemis version soon, because of the following issue:
https://issues.apache.org/jira/browse/ARTEMIS-3843

I would be interesting to know when a new release can be expected.

gsmet:
Fixes #76

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,3 +1,3 @@
 release:
-  current-version: 1.2.0
-  next-version: 1.2.1-SNAPSHOT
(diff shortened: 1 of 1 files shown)


gsmet:
@middagj FYI ^

I also changed the development version to `999-SNAPSHOT` as that's what we are doing these days.

Issue #36, opened 2022-04-12
Make quarkus.artemis.url optional


zhfeng:
I just moved https://github.com/quarkusio/quarkus/issues/18231 here to track it.

turing85:
@zhfeng What is the current status on how this should behave? Based on the implementation I am doing in #20, I can quite easily:
- return `null`s when the URL is not set at runtime
- exclude everything with `null`  URL from the health check(s).

This could ultimately lead to a situation where the health endpoint has no items, i.e. always returns `"UP"`.

Are those specifications acceptable?

PR #78, opened 2022-10-05
Feature/#20 add named connections

### LINKED ISSUE #20: Support for multiple named connection factories
Feature request similar to https://github.com/amqphub/quarkus-qpid-jms/issues/13

Support for multiple named connection factories similar to Quarkus datasources

    https://quarkus.io/guides/datasource#multiple-datasources

turing85:
Resolves #20


.all-contributorsrc
```diff
@@ -14,6 +14,15 @@
         "code",
         "maintenance"
       ]
+    },
+    {
+      in"login": "turing85",
```
middagj, line 19: `in` should be probably removed

core/common/src/main/java/io/quarkus/artemis/core/common/ArtemisUtil.java
```diff
@@ -0,0 +1,13 @@
+package io.quarkus.artemis.core.common;
+
+public class ArtemisUtil {
```
middagj, line 3: Any reason we cannot put this in `runtime`? It is inherited in all other modules.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreConfiguredBuildItem.java
```diff
@@ -1,9 +0,0 @@
-package io.quarkus.artemis.core.deployment;
-
-import io.quarkus.builder.item.SimpleBuildItem;
-
-/**
- * Marker build item indicating that Artemis Core is configured
- */
-public final class ArtemisCoreConfiguredBuildItem extends SimpleBuildItem {
```
middagj, line 8: Why remove it? It is a breaking change as some other libraries might depend on this to determine if they need to do extra work.

docs/modules/ROOT/pages/includes/quarkus-artemis-core.adoc
```diff
@@ -10,38 +10,22 @@ h|[[quarkus-artemis-core_configuration]]link:#quarkus-artemis-core_configuration
 h|Type
 h|Default
 
-a|icon:lock[title=Fixed at build time] [[quarkus-artemis-core_quarkus.artemis.health.enabled]]`link:#quarkus-artemis-core_quarkus.artemis.health.enabled[quarkus.artemis.health.enabled]`
```
middagj, line 13: Is it necessary to change config names? I thought we made it the same as other extensions, and it is a breaking change.

integration-tests/core/src/main/resources/application.properties
```diff
@@ -1 +1 @@
-quarkus.artemis.devservices.enabled=false
+quarkus.artemis.devservices.enabled=false
```
middagj, line 1: please don't reformat if not touched

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisRuntimeConfig.java
```diff
@@ -2,28 +2,44 @@
...
-@ConfigRoot(name = "artemis", phase = ConfigPhase.RUN_TIME)
+@ConfigGroup
 public class ArtemisRuntimeConfig {
 
     /**
      * Artemis connection url
      */
     @ConfigItem
-    public String url;
+    Optional<String> url;
```
middagj, line 16: url is not optional. By making it non-optional it saves you from checking later (as the only thing done is throw an exception) and if you make it non optional here, Quarkus will give a meaningful error.

Or is this needed for multiple connections?

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/health/ConnectionFactoryHealthCheck.java
```diff
@@ -10,20 +14,41 @@
...
 
     @Override
     public HealthCheckResponse call() {
-        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Artemis JMS health check");
-        try (Connection connection = connectionFactory.createConnection()) {
-            builder.up();
-        } catch (Exception e) {
-            builder.down();
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Artemis JMS health check").up();
+        for (Map.Entry<String, ConnectionFactory> namedConnectionFactory : connectionFactories.entrySet()) {
```
middagj, line 44: `var entry` would be better here,

just as:
`final ConnectionFactory connectionFactory` can be better readable as `var connectionFactory`. `final` on local variables isn't really necessary, but that can be left.

At least I thought we don't support Java 8 anymore.

Also final on local variables isn't really necessary, but you can

core/common/src/main/java/io/quarkus/artemis/core/common/ArtemisUtil.java
```diff
@@ -0,0 +1,13 @@
...
+    public static final String DEFAULT_CONFIG_NAME = "<default>";
+    public static final String DEFAULT_CONFIG_URL = "tcp://localhost:61616";
```
middagj, line 5: No config url should mean use devservices

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
```diff
@@ -2,112 +2,126 @@
...
     };
 
-    static final Class[] BUILTIN_LOADBALANCING_POLICIES = {
+    static final Class<?>[] BUILTIN_LOADBALANCING_POLICIES = {
             FirstElementConnectionLoadBalancingPolicy.class,
             RandomConnectionLoadBalancingPolicy.class,
             RandomStickyConnectionLoadBalancingPolicy.class,
             RoundRobinConnectionLoadBalancingPolicy.class
     };
+    protected static final String FEATURE = "artemis-core";
```
middagj, line 44: can be `private`

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
```diff
@@ -2,112 +2,126 @@
...
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
 
     @BuildStep
     NativeImageConfigBuildItem config() {
         return NativeImageConfigBuildItem.builder()
-                .addRuntimeInitializedClass("org.apache.activemq.artemis.api.core.ActiveMQBuffers")
```
middagj, line 49: I thought there was a reason to use strings, does this work?

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
```diff
@@ -2,112 +2,126 @@
...
+        addBuildintReflectiveBuiltitems(reflectiveClass, BUILTIN_CONNECTOR_FACTORIES);
 
-        for (ClassInfo ci : connectorFactories) {
-            LOGGER.debug("Adding reflective class " + ci);
-            reflectiveClass.produce(new ReflectiveClassBuildItem(false, false, ci.toString()));
-        }
+        final Collection<ClassInfo> loadBalancers = indexBuildItem.getIndex()
+                .getAllKnownImplementors(DotName.createSimple(ConnectionLoadBalancingPolicy.class.getName()));
+        addDynamicReflectiveBuildItems(reflectiveClass, loadBalancers);
+        addBuildintReflectiveBuiltitems(reflectiveClass, BUILTIN_LOADBALANCING_POLICIES);
```
middagj, line 70: `addBuildint` should be `addBuiltin`

middagj (commented):
I took a quick look and really appreciate taking the effort to implement this. Have some comments. If I have some time, let me check how to get devservices still working.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
turing85: We are still calling it with a `String`, since we call `...class.getCanonicalName()`

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/health/ConnectionFactoryHealthCheck.java
turing85: done

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
turing85: done

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
turing85: done

integration-tests/core/src/main/resources/application.properties
turing85: done

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreConfiguredBuildItem.java
turing85: done

docs/modules/ROOT/pages/includes/quarkus-artemis-core.adoc
turing85: I am still playing around with those options. Right now, it is challenging to determine at runtime which configurations exist since we do not have a configuration that must be present and is built-time fixated (like `db-kind` in datasources, for example).

.all-contributorsrc
turing85: done

core/common/src/main/java/io/quarkus/artemis/core/common/ArtemisUtil.java
turing85: nope, moved, and removed module

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisRuntimeConfig.java
turing85: It is in deed. Right now, the "root" configuration "is-a" runtime configuration, i.e. "inherits" all attributes. But this means that the url on the root should not be required. Furthermore, we most probably do not want a default connection as soon as we have one named connection (unless the default connection is explicitly configured). See https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisRuntimeConfigs.java#L15

turing85:
State so far:
As far as I can see, the connection factories are created correctly. Health endpoints show additional connections. Due to the missing integratoin with devservices, however, we have no tests on this part (yet).

Problematic: even if no "default" connection is configured, one is created. We would need to somehow be able to analyze the keys present at build-time. In particular, we would need to see whether any named connection is present and "suppress" the creation of a default connection if no keys (`url`, `username`, `password`, ...) are present. This is - to the best of my knowledge - currently not (easily) possible.

Another problem (bug?) is config-related. If we add the line

    quarkus.artemis."foo".health.enabled=true

then quarkus complains that the config key `quarkus.artemis."foo".health.health.enabled` is missing. This does not happen on configkey `devservices`. I suspect that is due to the fact that [the property `healthEnabled` is bound to the path `health.enabled`](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisBuildTimeConfig.java#L11).

turing85:
Integration with testcontainers is working properly, injection works as expected.

docs/modules/ROOT/pages/includes/quarkus-artemis-core.adoc
turing85: FTR: https://github.com/quarkusio/quarkus/issues/28430

core/common/src/main/java/io/quarkus/artemis/core/common/ArtemisUtil.java
turing85: Done

docs/modules/ROOT/pages/includes/quarkus-artemis-core.adoc
turing85: I aligned the config properties with the properties used by `quarkus.datasource`, i.e.:
- `quarkus.artemis.health.enabled` to enable health checking for artemis
- `quarkus.artemis.health-exclude` and `quarkus.artemis."named-connection".health-exclude` to exclude specific connections from health checking
- `quarkus.artemis.xa-enabled` and `quarkus.artemis."named-connection".xa-enabled` to activate XA transactions (beign in line with `quarkus.artemis.health-exclude`).

Sub-properties with dots in names, as described in the bug linked above, don't currently work. We can work around this by creating an object for `xa`, having a field `enable`. This would, however, be not in line with how `quarkus.datasource` does things.

turing85:
Implementation is done. As far as I can see, everything on the implementation-side works as expected.

Open points:
- I use [a hack](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ShadowRunTimeConfig.java) to get the runtime-configurations at build time in order to get the names of all connections. This hack is
  1. very dirty
  1. leads to problems wrt. documentation, entries are generated twice
  I have currently no workaround for this. As stated above, I have opened a feature request in quarkus ot resolive this.
 - Tests are still missing
 - With regards to tests, there is an open issue with [`ArtemisTestResources`](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/test-framework/src/main/java/io/quarkus/artemis/test/ArtemisTestResource.java): It only activates devservices for the default connection. I am if it is possible to activate **all** connections (and if so, how it would be done).

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
turing85: Can this be resolved?

docs/modules/ROOT/pages/includes/quarkus-artemis-core.adoc
middagj: Thanks for the clarification. Alignment with other extensions, especially Quarkus' is always good (my preference is also dash over camelCase). I guess, it is easily fixable on upgrade, we just need to do a minor version bump.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java
middagj: Yes, I think pipeline will break if it doesn't work.

middagj:
>     * I use [a hack](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ShadowRunTimeConfig.java) to get the runtime-configurations at build time in order to get the names of all connections. This hack is
>       
>       1. very dirty
>       2. leads to problems wrt. documentation, entries are generated twice
>          I have currently no workaround for this. As stated above, I have opened a feature request in quarkus ot resolive this.

Just curious, how do other extensions with multiple connections do this?

> 
>     * Tests are still missing
> 
>     * With regards to tests, there is an open issue with [`ArtemisTestResources`](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/test-framework/src/main/java/io/quarkus/artemis/test/ArtemisTestResource.java): It only activates devservices for the default connection. I am if it is possible to activate **all** connections (and if so, how it would be done).

I looked this week, but looks like we test both devservices and embedded. I think devservices only for default connection is an acceptable outcome. Can we then test the multiple connections only with the embedded? It can of course be to the same instance, perhaps we modify the instance to have two users.


Really great to see all the work done.

turing85:
> ....
> Just curious, how do other extensions with multiple connections do this?
> ...

`quarkus.datasource` and `mp.messaging` have both a required field (`db-type` for datasource, `connector` for messaging). We do not have such a field. The more I think about this, the more I come to the conclusion that everything messaging-related (jms, amqp, kafka, ...) should be grouped under a common configkey, just like 'quarkus.datasource`, with a discriminator-property that must be set.

> 
> ...
> I looked this week, but looks like we test both devservices and embedded. I think devservices only for default connection is an acceptable outcome. Can we then test the multiple connections only with the embedded? It can of course be to the same instance, perhaps we modify the instance to have two users.
> ...

Yes, we test both. I have not yet taken a deeper look on how the test extensions work specifically, I think that bot testing devservices and embedded should be possible, but with a custom test resource for `embedded`. Testing with devservices should already work.

turing85:
Adapting and extending the tests for `artemis-core` was straight forward. I think I should finish work on the `artemis-jms` tests today.

turing85:
integration tests are done & done. Merge Request is ready for review.

turing85:
If I understand it correctly, this MR should also close #36. Bu t I do not fully understand what the exact use case in camel is.

turing85:
I talked to @Ladicek today [in Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/.E2.9C.94.20ArC.20with.20named.20and.20unnamed.20beans/near/303210847). He recommends foregoing `@Named` and using `@Indentifier` instead. @middagj your opinion on this?

middagj:
> I talked to @Ladicek today [in Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/.E2.9C.94.20ArC.20with.20named.20and.20unnamed.20beans/near/303210847). He recommends foregoing named(...)/@Named ([here](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ArtemisCoreProcessor.java#L144) and [here](https://github.com/turing85/quarkus-artemis/blob/feature/20-add-named-connections/jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/ArtemisJmsProcessor.java#L104)) and using either @Indentifier or a custom qualifier annotation, like @Connection, instead. @middagj your opinion on this?

That thread is a good read (currently on a project back in Spring world, but I guess it is kind of the same).

I agree that `@Identifier` or a custom qualifier is a better option then `@Named`. I am a bit indifferent to which.

> EDIT: I have a stash ready that incorporates a new @Connection-annotation to replace @Named. So if we want to go down this route, just say the word.

Let's follow `DataSource` example and use our own. Should we ask on https://github.com/amqphub/quarkus-qpid-jms/issues/13 if we can agree to `@Identifier`. It will make life easier for users. Otherwise we can follow `DataSource` and use our own.

turing85:
As soon as the clarification from amqphub comes, I'll push my stash to introduce `@Connection`. The MR should then be considered finalized and ready to merge. I would, however, suggest to postpone the next release until #36 is implemented (I have already started working on that, but it might be necessary to move some properties from build- to runtime).

turing85:
amqphub clarified, they want to use `@Identfier`. I will align our implementation to do so as well.

turing85:
Done & ready for merge.

middagj (approved):

middagj:
Approval or MR is different from approval of workflow. Let's see if it finished and merge it.

middagj:
You can squash and rebase, otherwise I will do it via GitHub

turing85:
Pipeline is green, will rebase & squash now

turing85:
Done

middagj:
Thanks again!

PR #82, opened 2022-10-15
#36: make configuration value url optional

### LINKED ISSUE #36: Make quarkus.artemis.url optional

turing85:
Resolves #36


core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -193,17 +193,21 @@ private static RunningDevService startArtemis(
             DockerStatusBuildItem dockerStatusBuildItem,
             ArtemisDevServiceCfg config,
             LaunchModeBuildItem launchMode,
-            Optional<Duration> timeout) {
+            @SuppressWarnings("OptionalUsedAsFieldOrParameterType") Optional<Duration> timeout) {
```
middagj, line 196: Can we not set this at the class level, it is how Quarkus processors work.

middagj:
Pretty big, will review it later next week

turing85:
It looks actually bigger than it is. Most changes are due to additional test modules, but the tests are mostly the same, with only minor changes.

I also took the liberty and did some general cleanup, e.g. replacing field- with constructor-injection where possible, some renamings for clarity, ...

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
turing85: Done. For the extension part, I have moved this suppress up to all class-levels in all extension-modules.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisCoreRecorder.java
```diff
@@ -16,17 +17,27 @@ public Supplier<ServerLocator> getServerLocatorSupplier(
...
         ArtemisBuildTimeConfig buildTimeConfig = buildTimeConfigs.getAllConfigs().getOrDefault(name,
                 new ArtemisBuildTimeConfig());
-        ArtemisUtil.validateIntegrity(runtimeConfig, buildTimeConfig, name);
+        ArtemisUtil.validateIntegrity(name, runtimeConfig, buildTimeConfig);
+        ServerLocator serverLocator = getServerLocator(runtimeConfig);
         return new Supplier<>() {
             @Override
             public ServerLocator get() {
-                try {
-                    return ActiveMQClient.createServerLocator(runtimeConfig.getUrl());
```
middagj, line 24: Is there no side effect of creating the ServerLocater earlier than the call to `Supplier.get` ?

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
```diff
@@ -85,26 +87,37 @@ public int getPort() {
     }
 
     public String getImageName() {
-        return imageName;
+        return imageName.orElse("quay.io/artemiscloud/activemq-artemis-broker:1.0.5");
```
middagj, line 90: `1.0.9`

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
```diff
@@ -85,26 +87,37 @@ public int getPort() {
...
-        return password;
+        return password.orElse("admin");
     }
 
     public String getExtraArgs() {
-        return extraArgs;
+        return extraArgs.orElse("--no-autotune --mapped --no-fsync");
+    }
+
+    public boolean isEmpty() {
```
middagj, line 113: I don't know if I like all those `isEmpty` checks aggregated, very easy to forget one. Do we really need to be optional here? The issue only mentions `url`, I guess it also holds for `xaEnabled`, but I don't see why we need to add this to DevServices as well.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/health/ServerLocatorHealthCheck.java
```diff
@@ -14,34 +14,36 @@
...
+    public ServerLocatorHealthCheck(
+            @SuppressWarnings("CdiInjectionPointsInspection") ArtemisRuntimeConfigs runtimeConfigs,
+            @SuppressWarnings("CdiInjectionPointsInspection") ArtemisHealthSupport support) {
         Set<String> names = support.getConfiguredNames();
         Set<String> excludedNames = support.getExcludedNames();
         for (String name : names) {
-            ServerLocator locator = Arc.container().instance(ServerLocator.class, Identifier.Literal.of(name)).get();
-            if (!excludedNames.contains(name) && locator != null) {
-                serverLocators.put(name, locator);
+            if (runtimeConfigs.getAllConfigs().getOrDefault(name, new ArtemisRuntimeConfig()).getUrl() != null) {
```
middagj, line 32: Can `name` not be found in `runtimeConfigs`? We now create a new instance for every name in every health check.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/health/ServerLocatorHealthCheck.java
```diff
@@ -14,34 +14,36 @@
...
+                ServerLocator locator = Arc.container().instance(ServerLocator.class, Identifier.Literal.of(name)).get();
+                if (!excludedNames.contains(name) && locator != null) {
```
middagj, line 34: Why find the locator if it would be excluded?

middagj (commented):
I checked the source code and put a few comments and skimmed the tests. If there is anything of particular interest in the tests, let me know.

@ppalaga Can you check if this would solve your issue?

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
turing85: The `isEmpty()` - as-is - is the only way I was able to detect reliably what configuration(s) are present at compile-time. For this, we also need to access the runtime properties to detect whether anything is present. This is necessary since we have to call the recorder once per configuration.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/health/ServerLocatorHealthCheck.java
turing85: The name can, the URL cannot. I can wrap the call in an `Optional` and call `.orElseGet(ArtemisRuntimeConfig::new)`.

turing85:
> I checked the source code and put a few comments and skimmed the tests. If there is anything of particular interest in the tests, let me know.

Only important question is: is the amount of tests (and especially the amount of native image builds) okay? Right now, we are building six native images, a build runs for ~30 minutes.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/health/ServerLocatorHealthCheck.java
turing85: I have refactored this part to not check the URL. If the configuration is active, and the URL is not set, then the call to ÀrtemisUtil::validateIntegrity` will throw an exception.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/health/ServerLocatorHealthCheck.java
turing85: fixed.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
turing85: fixed

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisCoreRecorder.java
turing85: Refactored to create the locator in the supplier. Moved the integrity check also down to the supplier.

turing85: Refactored to create the locator in the supplier. Moved the integrity check also down to the supplier.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
middagj: I see, let's leave it like this for now.

middagj:
> Only important question is: is the amount of tests (and especially the amount of native image builds) okay? Right now, we are building six native images, a build runs for ~30 minutes.

That is really the downside of native builds. I don't know if there is an option to only do it once discussion is over.

middagj:
Thanks again! I leave it open to get feedback from Camel people, but looks good to me.

zhfeng:
Thanks @turing85 and I think it is good to resolve https://github.com/apache/camel-quarkus/issues/2857.
So if there is a empty config, no `ConnectionFactory` bean will be generated and `healthcheck` will return 'UP'. Is it right? Aslo I think we need to update https://github.com/quarkiverse/quarkus-artemis/blob/main/docs/modules/ROOT/pages/index.adoc to add a section to explian the multiple named connection factory configuration and some notes.

turing85:
> Thanks @turing85 and I think it is good to resolve [apache/camel-quarkus#2857](https://github.com/apache/camel-quarkus/issues/2857). So if there is a empty config, no `ConnectionFactory` bean will be generated and `healthcheck` will return 'UP'. Is it right? 

This is partially correct. Basically, if we load the extension without any configuration, no injectable bean will be created, and the health endpoint will return `UP`, but the corresponding health check will not be loaded (see [here](https://github.com/turing85/quarkus-artemis/blob/feature/36-make-url-optional/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/health/ArtemisHealthProcessor.java#L38)). If we still want the health check to be loaded, then this should be an easy fix. But the check will not pick up the beans produced outside the extension. It would be possible to pick them up, however, we would have no chance of giving them meaningful names.

zhfeng:
OK, I get it. In camel-qurakus-jms, we have a use case to create a custom `ConnectionFactory` such like
```
@Named("custom")
ConnectionFatory createCustomConnectionFactory() {
...
}
```
and to use it in the camel routers. So if we can find a way to enable the health check in this scenario, it would be great. I'm not sure what is about the `meaningful names`?

turing85:
The health of those connection factories are represented as key-value pairs in the health check. The key is a display name, the value is it status. In the example you gave, the name is `"custom"` (I'd btw recommend to not use `@Named`, but other annotations like `@Identifier` or a custom scope). It would be possible to get all instances of the `ConnectionFactory`-interface, but we are not (easily) able to detect the annotation that qualifies the name.

zhfeng:
Thanks for explainations. Can we create a follow up issue to track this improvement? and all of others are good to me. Thanks for your contribution again!

zhfeng (approved):

turing85:
@zhfeng I have created #84 to track this issue.

On a sidenote: I took another look at the source code of the pooled-jms extension, in particular [`PooledJmsWrapper`, line 44](https://github.com/quarkiverse/quarkus-pooled-jms/blob/main/runtime/src/main/java/io/quarkiverse/messaginghub/pooled/jms/PooledJmsWrapper.java#L44). Am I correct that, for XA-enabled connection factories, connection pooling is not enabled?

zhfeng:
No, I think the pooling is still enabled and just add support for transaction manager integration.

turing85:
Yeah, I saw the call to `pooledJmsRuntimeConfigureConnectionFactory(...)` after opening the code in an IDE :) As I said: Brain fart

PR #86, opened 2022-10-25
#84: add beans created outside the extension to health check

### LINKED ISSUE #84: Health check: check ServerLocator- and ConnectionFactory-beans created outside the extension
This is a follow-up of #36.

As of now, `ServerLocator`- and `ConnectionFactory`-beans created outside the extension. In general. however, this should be possible by:
- getting all beans implementing `ServerLocator`- or `ConnectionFactory`-beans,
- filtering-out the ones already handled by the extension, and
- add the beans to the health check.
We will attribute those beans created outside the extension as *external bean* for the following disucssion.

Challenges:
1. We need to generate sensible names for the external beans. We cannot reliably detect scope-annotations since there is a multitude of annotations that could be used, including user-defined scopes.
2. There might be an issuse with "duplicate" beans. IIRC, the [quarkus-pooled-jms extension](https://github.com/quarkiverse/quarkus-pooled-jms) wraps already-existing connection factories into a wrapper-factory for pooling. Thus, we have - on a technical level - two `ConnectionFactory`-beans representing the same connection.

turing85:
Resolves #84


zhfeng (approved):
Thanks @turing85 and it looks good to me!

Awesome works!

middagj (approved):

zhfeng:
@middagj can we do a new release to include these greate works recently?

middagj:
> @middagj can we do a new release to include these greate works recently?

Yes we can.

PR #92, opened 2022-10-28
#91: improved documentation

### LINKED ISSUE #91: Improve documentation
To facilitate the usage of the new features, we should improve our documentation.

turing85:
Resolves #91

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -43,10 +43,241 @@ Once the BOM is added, simply add the `io.quarkiverse.artemis:quarkus-artemis-jm
     <groupId>io.quarkiverse.artemis</groupId>
     <artifactId>quarkus-artemis-jms</artifactId>
 </dependency>
+<dependency>
+    <groupId>io.quarkiverse.artemis</groupId>
+    <artifactId>quarkus-test-artemis</artifactId>
+    <scope>test</scope>
+</dependency>
 ----
 
+We recommend to set properties `quarkus.version` and `artemis.version` to the versions you are using/need to use. We also recommend to align `artemis.version` with the artemis server version used.
+
+== Named configurations
+As with `quarkus.datasource`, we can configure multiple, named configuration. If a configuration is configured under the root `quarkus.artemis`, then this is the default or unnamed configuration. This corresponding bean is registered as the `@Default` bean and can be injected without additional qualifiers.
+
+.Injection of default `ServerLocator` when using `quarkus-artemis-core`
+[source,java,subs=attributes+]
+----
+public class MyClass {
+    ...
+    private final ServerLocator serverLocator;
+    ...
+
+    public MyClass (
+            ...
+            ServerLocator serverLocator,
+            ...
+    ) {
+        ...
+        this.serverLocator = serverLocator;
+        ...
+    }
+}
+----
+
+.Injection of default `ConnectionFactory` when using `quarkus-artemis-jms`
+[source,java,subs=attributes+]
+----
+public class MyClass {
+    ...
+    private final ConnectionFactory connectionFactory;
+    ...
+
+    public MyClass (
+            ...
+            ConnectionFactory connectionFactory,
+            ...
+    ) {
+        ...
+        this.connectionFactory = connectionFactory;
+        ...
+    }
+}
+----
+
+It is also registered with `@Identifier("<default>")`, so we can inject it with
+
+.Injection of default `ServerLocator` by its explicit name when using `quarkus-artemis-core`
+[source,java,subs=attributes+]
+----
+public class MyClass {
+    ...
+    private final ServerLocator serverLocator;
+    ...
+
+    public MyClass (
+            ...
+            @Identifier("<default>") ServerLocator serverLocator,
+            ...
+    ) {
+        ...
+        this.serverLocator = serverLocator;
+        ...
+    }
+}
+----
+
+.Injection of default `ConnectionFactory` by its explicit name when using `quarkus-artemis-jms`
+[source,java,subs=attributes+]
+----
+public class MyClass {
+    ...
+    private final ConnectionFactory connectionFactory;
+    ...
+
+    public MyClass (
+            ...
+            @Identifier("<default>") ConnectionFactory connectionFactory,
+            ...
+    ) {
+        ...
+        this.connectionFactory = connectionFactory;
+        ...
+    }
+}
+----
+
+Named configurations, on the other hand, are not configured as `@Default` bean and must therefore be injected by name. The name of the bean is the name of the configuration, and `@Identifier(...)` is used to qualify the name. So for example, if we define a configuration "named-1":
+
+.Named configuration "named-1"
+[source,properties,subs=attributes+]
+----
+quarkus.artemis."named-1".url=...
+quarkus.artemis."named-1".username=...
+quarkus.artemis."named-1".password=...
+----
+
+we can inject it with:
+
+.Injection of named `ServerLocator` "named-1" when using extension `quarkus-artemis-core`
+[source,java,subs=attributes+]
+----
+public class MyClass {
+    ...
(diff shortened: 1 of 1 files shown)


docs/modules/ROOT/pages/index.adoc
```diff
@@ -43,10 +43,236 @@ Once the BOM is added, simply add the `io.quarkiverse.artemis:quarkus-artemis-jm
...
+----
+
+We can then annotate our test with `@QuarkusTestResource(NamedOneArtemisTestResource.class)`.
+
+This instance can be configured by placing a `broker-named-1.xml` in `src/test/resources`. In general, for a configuration with name `<name>`, we can configure the corresponding in-memory artemis through file `src/test/resources/broker-<name>.xml`.
+
+For an in-depth explanation of what can be configured in a `broker.xml`, please see the link:https://activemq.apache.org/components/artemis/documentation/latest/configuration-index.html:[official Artemis documentation].
+
+== Examples
+Examples can be found in the `integration-tests` module, e.g link:https://github.com/quarkiverse/quarkus-artemis/tree/main/integration-tests/core/with-default/src/test/resources[The Apache ActiveMQ Artemis Core integration with default configuration tests module]
```
turing85, line 275: @middagj  @zhfeng 
I'd like to have a review of this part in particular. I am sure that there must be a better way to link to a project resource.

turing85:
When this PR is approved and merged, I'll come up with some release notes to version '2.0.0`.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -43,10 +43,236 @@ Once the BOM is added, simply add the `io.quarkiverse.artemis:quarkus-artemis-jm
...
+        this.connectionFactory = connectionFactory;
+        ...
+    }
+}
+----
+
+IMPORTANT: We strongly recommend the usage of link:https://javadoc.io/doc/io.smallrye.common/smallrye-common-annotation/latest/io/smallrye/common/annotation/Identifier.html[`@Identifier` annotation] instead of the link:https://www.javadoc.io/doc/javax/javaee-api/latest/javax/inject/Named.html[`@Named` annotation]. Some other feature rely on the usage of `@Identfier`. Please see section <<_health_checks>> for details.
+
+== Configuration detection at build-time
+We took special care so the configurations behave "as intuitively as possible". This means that if no connection-related configuration (`enabled`, `url`, `useranme`, `password`, `devservices...`, `health-exclude`, `xa-enabled`) is present at build-time, it is assumed that the  configuration is not used and disabled. Therefore, if we want to use any configuration, but not configure it, we should set `quarkus.artemis.enabled` / `quarkus.artemis."named-configuration".enabled` to *true* to explicitly enable the configuration.
```
middagj, line 193: typo in `username`

middagj, line 193: I understand what is being said here, but I don't understand the why. What would be the use case to leave out all configurations, but do enable the config 😕

docs/modules/ROOT/pages/index.adoc
middagj: I think a relative reference should work: https://github.blog/2013-01-31-relative-links-in-markup-files/

middagj:
> Version Updates:
>  - Upgraded quarkus from 2.13.0.FINAL to 2.13.2.FINAL

I would not use the work _Upgraded_ as it implies that we update our versions when Quarkus updates. I would only mention compatibility issues if we needed to change something to make our library work for a new Quarkus version.

middagj (approved):
Some small things

docs/modules/ROOT/pages/index.adoc
turing85: Camel. At least that's how I interpreted a corner-case of ["*Camel has some alternative ways to configure Artemis, e.g. end users can produce their own ConnectionFactory.*"](https://github.com/quarkusio/quarkus/issues/18231).

turing85:
> > Version Updates:
> > 
> > * Upgraded quarkus from 2.13.0.FINAL to 2.13.2.FINAL
> 
> I would not use the work _Upgraded_ as it implies that we update our versions when Quarkus updates. I would only mention compatibility issues if we needed to change something to make our library work for a new Quarkus version.

Hmm.. instead of "Upgraded" just "Changed"?

docs/modules/ROOT/pages/index.adoc
turing85: hm so...  `file:///../../../../integration-tests/core/with-default/src/test[The Apache ActiveMQ Artemis Core integration with default configuration tests module]` should work?

middagj: I thought more like:

```adoc
file://integration-tests/core/with-default/src/test[The Apache ActiveMQ Artemis Core integration with default configuration tests module]
```

Problem here is the feedback loop of 6 hours?

turing85: ... yes. Let's embrace agile: move fast, break things :slightly_smiling_face:

docs/modules/ROOT/pages/index.adoc
middagj: Then I would change it like:

```
We took special care so the configurations behave "as intuitively as possible". In general this means that if no connection-related configuration fields (`enabled`, `url`, `username`, `password`, `devservices...`, `health-exclude`, `xa-enabled`) are present at build-time it is disabled. If you produce your own ConnectionFactory you can set `enabled` to `true` to explicitly enable the configuration.```

But I don't understand when you need this, because as far as I understand now, is that they even produce their own default ConnectionFactory. So they don't need to configure one via our extension because `url` is optional now.

turing85: Through `@Identtifier`, they get helath checks for free.

middagj:
> Hmm.. instead of "Upgraded quarkus..." we could write "Changed quakrus version...". Would this be more suitable?

How about. "Confirmed compatibility with Quarkus up to version ..." ?

But changed is also better.

turing85:
> > Hmm.. instead of "Upgraded quarkus..." we could write "Changed quakrus version...". Would this be more suitable?
> 
> How about. "Confirmed compatibility with Quarkus up to version ..." ?
> 
> But changed is also better.

I tend to favor "Changed" since we had a change in our repository. "Confirmed" has a passive nature to it.

docs/modules/ROOT/pages/index.adoc
middagj: Ah yes.

> We took special care so the configurations behave "as intuitively as possible". In general this means that if no connection-related configuration fields (enabled, url, username, password, devservices..., health-exclude, xa-enabled) are present at build-time it is disabled. If you produce your own ConnectionFactory you can set enabled to true to explicitly enable the configuration to include it in health-checks.

Like this?

middagj: Or your own wording, but adding the example or use case makes it understandable.

turing85: >  ...
> > ...If you produce your own ConnectionFactory you can set `enabled` to `true` to explicitly enable the configuration.
>  ...

This is actually not correct. Our extension does nothing with beans provided in code (no wrapping, nothing). The only "service" we provide, as stated above, is inclusion in health checks when the bean is annotated with `@Identifier(...)`.

middagj: But how do they use `enabled then?

turing85: Why should they? The `ConnectionFactory` is provided in-code. We cannot enalbe/disalbe this bean since its creation is out of our control. Or am I misunderstanding something?

turing85: FTR: https://quarkusio.zulipchat.com/#narrow/pm-with/285062,403386-pm/near/306979669

turing85: fixed.

zhfeng (approved):

PR #101, opened 2022-11-03
Feature/90 fix property bind at runtime

### LINKED ISSUE #90: Question: Release notes
Hello,

I don't find any documentation about changes in the releases. Are there any release notes?
For us it is important to know for example when the Artemis version was changed (which of course can be found in the pom.xml) and of course when there are breaking changes. For example as far as I understood the release version 2.0.0 was created, because of changes in the config. But what are those changes?
After I switched to the latest version, our tests no longer work. Now I wonder if there is a change I could/should have noticed via release notes.

turing85:
Fixes #90


core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -279,15 +281,17 @@ private static final class ArtemisDevServiceCfg {
...
-            this.fixedExposedPort = config.getPort();
-            this.shared = config.isShared();
-            this.serviceName = config.getServiceName() + "-" + name;
-            this.user = config.getUser();
-            this.password = config.getPassword();
-            this.extraArgs = config.getExtraArgs();
+        public ArtemisDevServiceCfg(ArtemisBuildTimeConfig config, ShadowRuntimeConfig shadowRuntimeConfig, String name) {
+            ArtemisDevServicesBuildTimeConfig devServicesConfig = config.getDevservices();
+            this.devServicesEnabled = devServicesConfig.enabled
+                    .orElseGet(() -> shadowRuntimeConfig.url.isEmpty() ? true : false);
```
middagj, line 287:
```java
.orElse(shadowRuntimeConfig.url.isEmpty())
```

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ShadowRunTimeConfigs.java
```diff
@@ -0,0 +1,48 @@
+package io.quarkus.artemis.core.deployment;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import io.quarkus.artemis.core.runtime.ArtemisUtil;
+import io.quarkus.runtime.annotations.*;
+
```
zhfeng, line 10: Can we leave a comment here to mention that it is only used for reading the keys? and it could be improved by https://github.com/quarkusio/quarkus/issues/28415.

middagj (approved):
Minor point

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ShadowRunTimeConfigs.java
turing85: Yes. I will also try to restructure those classes so it is obvious we access only keys and query presence of values, to furher clarify the use case for those classes.

turing85: Done. 
@zhfeng care to take another look?

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
turing85: Done.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ShadowRunTimeConfigs.java
zhfeng: yeah, it looks good to me. Thanks again!

zhfeng (approved):

turing85:
Thanks for the review. I'll prepare the release for version `2.0.1` and write release notes as soon as the pipeline succeeds.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/ShadowRunTimeConfigs.java
zhfeng: Hmm, it seems that I have no permission to resolve the conversation. Anyway, please go head for a new release.

turing85: As soon as the pipeline succeeds (the amount of tests we have, and the fact that all tests also compile natively, means that the build takes quite some time).

PR #119, opened 2022-12-07
Rewrite get-date pipeline step to remove warning

### LINKED ISSUE #118: Pipeline step "Get Date" produces warning
In our pipeline-runs, the step `get-date` produces a warning:

```
Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

See e.g. [here](https://github.com/quarkiverse/quarkus-artemis/actions/runs/3636261517/jobs/6136072642#step:4:7). 

We should rewrite this part to remove the warning.

turing85:
- Fixes #118

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up JDK 11
         uses: actions/setup-java@v3
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,22 +31,11 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
-          distribution: temurin
           java-version: 11
-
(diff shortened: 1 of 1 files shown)


gastaldi:
I'd be in favor of removing this step and use `cache: 'maven`` in the `actions/setup-java@v3` step. See https://github.com/quarkiverse/quarkus-prettytime/blob/main/.github/workflows/build.yml

turing85:
> I'd be in favor of removing this step and use `cache: 'maven'` in the `actions/setup-java@v3` step. See https://github.com/quarkiverse/quarkus-prettytime/blob/main/.github/workflows/build.yml

Do we have an automatic way to prevent the cache from growing over time, or is this just a no-concern?

gastaldi:
The cache is invalidated when the pom.xml changes, so that shouldn't be a problem after a release

gastaldi (approved):

.github/workflows/build.yml
```diff
@@ -31,23 +31,11 @@ jobs:
       - uses: actions/checkout@v2
```
gastaldi, line 31:
```suggestion
      - uses: actions/checkout@v3
```

Please bump `actions/checkout` too

PR #122, opened 2022-12-10
Better integration with camel

turing85:
Resolves #121

PR #125, opened 2023-01-03
Fix camel-jms tests

zhfeng:


jms/deployment/src/main/java/io/quarkus/artemis/jms/deployment/ArtemisJmsProcessor.java
```diff
@@ -95,6 +95,7 @@ private static SyntheticBeanBuildItem toSyntheticBeanBuildItem(
         SyntheticBeanBuildItem.ExtendedBeanConfigurator configurator = initializeConfigurator(isXaEnable)
                 .supplier(connectionFactorySupplier)
                 .scope(ApplicationScoped.class);
+        configurator.name(name);
```
turing85, line 98: We can just chain the call: 
```
initializeConfigurator(isXaEnable)
                .supplier(connectionFactorySupplier)
                .scope(ApplicationScoped.class)
                .name(name);
```

integration-tests/camel-jms/with-named-and-external/src/main/java/io/quarkus/it/artemis/camel/jms/withnamedandexternal/BeanProducer.java
```diff
@@ -17,6 +18,7 @@ public class BeanProducer {
     @Typed({ ConnectionFactory.class })
     @ApplicationScoped
     @Identifier("externally-defined")
+    @Named("externally-defined")
```
turing85, line 21: I don't like the fact that the externally defined connection factory needs the additional `@Named(...)` annotation. One goal of this feature for me is/was that `@Named(...)` is not necessary, the `CamelContextEnhancer` took care of this. 

From the point of view of users, it is hard to understand why they should add two annotations to the bean, definen the same name. Do we have other possibilities to - on a global level - tell camel that all `@Identifier(...)` beans should be part of the camel context? Is there a specific reason why we should remove the `CamelContextEnhancer`?

turing85 (changes requested):

integration-tests/camel-jms/with-named-and-external/src/main/java/io/quarkus/it/artemis/camel/jms/withnamedandexternal/BeanProducer.java
turing85: Hm... I'll test the water with `camel-quarkus` on this feature. Do you see any harm in keeping the `CamelContextEnhancer` until then?

zhfeng: No, if we decide to keep `CamelContextEnhancer` until we find a better way to do with `@Identifier` beans, I think it should be good to introduce a config property such like `quarkus.artemis.camel-quarkus-enhance-enabled` with default value `false` to protect this feature. And explain it in the document, if users define a external ConnectionFactory bean with `@Identifier`, it should enable this propery to make it availiable in the camel router definition unless add `@Named`. WDYT?

turing85: Sounds good to me. Should we close this PR, and I'll take care of enabling the context enhancer through a property?

zhfeng: Yeah, I think so and please keep in mind to add `configurator.name(name);` in `ArtemisJmsProcessor`. Thanks!

turing85:
closed in favor of a toggle for the `CamelContextEnhancer`; see https://github.com/quarkiverse/quarkus-artemis/pull/122/commits/77bb9a2d9d47dd0e0b8937fd1a9e667ed840b82c on #122. The additional `name(...)` were added in https://github.com/quarkiverse/quarkus-artemis/pull/122/commits/a4715297125215d60329f21e35f09ab80d8995d6 on #122.

PR #149, opened 2023-02-10
Fix #148 to use ConnectionFactoryWrapperItem

### LINKED ISSUE #148: Use ConnectionFactoryWrapperBuildItem in Quarkus
We need to use `ConnectionFactoryWrapperBuildItem` instead of  `ArtemisJmsWrapperBuildItem` since https://github.com/quarkusio/quarkus/pull/29753 get merged.

And I also create
- https://github.com/quarkiverse/quarkus-pooled-jms/issues/71

zhfeng:


zhfeng:
@turing85 can you take a review?

build-parent/pom.xml
```diff
@@ -35,7 +35,12 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-jms-spi-deployment</artifactId>
+                <version>${quarkus.version}</version>
```
turing85, line 41: Do we already have a PR in quarkus to fix this? If so: Can we add a TODO to remove this entry?

turing85 (changes requested):

build-parent/pom.xml
turing85: Relevant PR for `2.16`: https://github.com/quarkusio/quarkus/pull/31075

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/ArtemisJmsRecorder.java
```diff
@@ -13,20 +14,15 @@
...
 
-    public ArtemisJmsWrapper getDefaultWrapper() {
-        return new ArtemisJmsWrapper() {
-            @Override
-            public ConnectionFactory wrapConnectionFactory(ActiveMQConnectionFactory cf) {
-                return cf;
-            }
-        };
+    public Function<ConnectionFactory, Object> getDefaultWrapper() {
+        return cf -> cf;
```
middagj, line 18: I thought Quarkus has a policy of using anonymous classes instead of lambda because of performance reasons.

turing85: @middagj See [this zulip chat](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/lambdas.20in.20extensions). So we can use lambdas.

middagj: Great, I like the lambda notation better. I was planning to do the research myself as I mentioned it before (after got corrected on in https://github.com/quarkusio/quarkus/pull/15350#discussion_r583563115).

Makes sense, that the JVM optimizes these things by itself instead of relying on devs (same as with StringBuilder vs concat operator previously (nowadays it is even faster using invokedynamic)).

turing85: @middagj can this thread then be resolved?

build-parent/pom.xml
turing85: @zhfeng can you add the `TODO` in the pom? As far as I am concerned, this is the last blocker on the PR. I'd like to get this one merged before we switch the main to `3.0.0`-based implementation.

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/ArtemisJmsRecorder.java
middagj: Yes, of course

build-parent/pom.xml
zhfeng: Yeah, I can add it.

turing85 (approved):

PR #227, opened 2023-06-08
Fix all asciidoc warnings by using pass macro

TheJavaGuy:
Asciidoc interprets strings like ${ARTEMIS_NAMED_URL} as attributes and then issues a WARN during the build that it will skip reference to a missing attribute. This PR uses inline pass macro to tell asciidoc that enclosed content should be excluded from all substitutions.

Proof of all warnings gone:
image

Proof of warnings before this PR:
image

### DIFF
--- a/docs/modules/ROOT/pages/index.adoc
+++ b/docs/modules/ROOT/pages/index.adoc
@@ -253,8 +253,8 @@ a|
 a|
 configuration property bound to environment variable
 a|
-* `quarkus.artemis."named".url=${ARTEMIS_NAMED_URL}`
-* `quarkus.artemis.password=${ARTEMIS_PASSWORD}`
+* `quarkus.artemis."named".url=pass:[${ARTEMIS_NAMED_URL}]`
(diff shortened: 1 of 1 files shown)


turing85:
@TheJavaGuy thanks for the PR!

TheJavaGuy:
@turing85 Luckily, `main` branch doesn't have this problem, only `2.x` does.

turing85:
> @turing85 Luckily, `main` branch doesn't have this problem, only `2.x` does.

That's strange. Shouldn't [this line](https://github.com/quarkiverse/quarkus-artemis/blob/main/docs/modules/ROOT/pages/index.adoc?plain=1#L205) cause the same warning? Or does the colon prevent the warning?

TheJavaGuy:
@turing85 It looks like colon makes asciidoctor not to consider `${ARTEMIS_URL:tcp://dummy:12345}` to be a reference to an attribute. If it was just `${ARTEMIS_URL}` then we would have a warning there too.

PR #246, opened 2023-07-14
Use `graal-sdk` instead of `svm`

gastaldi:
- Caused by https://github.com/quarkusio/quarkus/pull/34145

### DIFF
--- a/core/runtime/pom.xml
+++ b/core/runtime/pom.xml
@@ -71,8 +71,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.graalvm.nativeimage</groupId>
-            <artifactId>svm</artifactId>
+            <groupId>org.graalvm.sdk</groupId>
+            <artifactId>graal-sdk</artifactId>
             <scope>provided</scope>
         </dependency>
     </dependencies>


turing85 (approved):

turing85:
@gastaldi As far as I can see, this change should not be downstreamed to the `2.x` versions, right?

gastaldi:
Because `org.graalvm.sdk:graal-sdk` is already in (always has been) in the Quarkus application's BOM, I don't believe there is a need to have a separate branch. 

It is a breaking change in the sense that the dependency is no longer available in the BOM, but it doesn't require changing any sources.

turing85:
@gastaldi thanks for the clarification :slightly_smiling_face:

turing85:
Final question: (when) will `svm` be removed? I.e. when do we need to release a new version?

maxandersen:
> Do we need a separate `3.0.x` branch to support older versions?

you most likely will unless you don't use any jakarta/javax dependencies ?

maxandersen:
@turing85 in 3.2.1 is the plan. next week :/

turing85:
> > Do we need a separate `3.0.x` branch to support older versions?
> 
> you most likely will unless you don't use any jakarta/javax dependencies ?

What does jakarta have to do with the svm dependency?
If you mean the switch from JavaEE to JakartaEE: we already migrated; that's what the `2.x` branch is for.

maxandersen:
@turing85 

> What does jakarta have to do with the svm dependency?
> If you mean the switch from JavaEE to JakartaEE: we already migrated; that's what the `2.x` branch is for.

sorry - I thought you mean Quarkus 2.0 vs 3.0 - if already branched you should be fine.

gemmellr:
I think the question was misunderstood. The graal-sdk dep has been in the bom for ages ([e.g 3.0.0](https://github.com/quarkusio/quarkus/blob/3.0/bom/application/pom.xml#L92)) so as George said, it doesnt seem like you should need a branch for older 3.x versions.

I've made the switch and released quarkus-qpid-jms against Quarkus 3.2 again without issue.

gemmellr:
> @turing85 in 3.2.1 is the plan. next week :/

Is it? The PR isnt marked for backport and prior discussion suggested it would wait for 3.3

turing85:
Should we move the discussion to the mailing list? I imagine there are other quarkiverse extensions that are affected as well.

gastaldi:
Yeah, I should have done that before opening the PRs, sorry. I'll start a thread in the quarkus-dev list

PR #322, opened 2023-11-03
Cleanup integration tests

turing85:
- When no message can be received, return null
- Where possible, rely on RestAssured


middagj (approved):
What is up with the formatter?

turing85:
> What is up with the formatter?

Do you mean the `// @formatter:off`? If you take it out, the formatter is of the oppinion that all of this should go on one line. Not really readable. I find that the test flow is much more readable with those line breaks

zhfeng (approved):

PR #348, opened 2023-12-01
Fix #347 to add commons-logging-jboss-logging

### LINKED ISSUE #347: Artemis native build is failing with java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
It seems that we remove the following in `core/runtime/pom.xml`
```
<dependency>
    <groupId>org.jboss.logging</groupId>
    <artifactId>commons-logging-jboss-logging</artifactId>
</dependency>
```
 in https://github.com/quarkiverse/quarkus-artemis/commit/d3b2cf0345d30eae4d2a1fc7bf4d42dcf881a12d#diff-40752b74da493b2aa06facf4b1ebd7a81736f6318dcfe76ddae90deafa57fad7

We have to add this artifact because it is needed by `artemis-core-client`.

The CI is working because `commons-logging-jboss-logging` is comming from `io.quarkus:quarkus-junit5`. We set it as `compile` scope in `integration-tests/core/common/pom.xml` and I think it coud be wrong.

zhfeng:
Fix #347

### DIFF
--- a/core/runtime/pom.xml
+++ b/core/runtime/pom.xml
@@ -61,6 +61,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.jboss.logging</groupId>
+            <artifactId>commons-logging-jboss-logging</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.graalvm.sdk</groupId>
             <artifactId>graal-sdk</artifactId>
             <scope>provided</scope>
--- a/integration-tests/camel-jms/common/pom.xml
+++ b/integration-tests/camel-jms/common/pom.xml
@@ -27,17 +27,4 @@
             </plugin>
         </plugins>
     </build>
-
-    <dependencies>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-junit5</artifactId>
-            <scope>compile</scope>
-        </dependency>
-        <dependency>
-            <groupId>io.rest-assured</groupId>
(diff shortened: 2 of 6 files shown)


core/runtime/pom.xml
```diff
@@ -60,6 +60,11 @@
             <artifactId>artemis-core-client</artifactId>
         </dependency>
 
+        <dependency>
```
turing85, line 63: Can we add a comment explicitly stating that we need this dependency for `artemis-core-client`?

I highly suspect that we need this dpenendency in [`jms/runtime/pom.xml`](https://github.com/quarkiverse/quarkus-artemis/blob/0d9a2ab0751707d28a0ac41948b870b733f46fce/jms/runtime/pom.xml) as well.

turing85 (changes requested):

core/runtime/pom.xml
turing85: Okay, the dependency is not needed in [`jms/runtime/pom.xml`](https://github.com/quarkiverse/quarkus-artemis/blob/0d9a2ab0751707d28a0ac41948b870b733f46fce/jms/runtime/pom.xml); the `jms/runtime` depends on the `core/runtime`, so this change is "propagated".

turing85: Nope :slightly_smiling_face:

turing85: @zhfeng did you forget to add the comment?

PR #410, opened 2024-02-08
Update github action "upload-artifact" from v3 to v4

turing85:


turing85:
This is more complex. We probably need a small rework of the pipleines. Closing for now.

PR #414, opened 2024-02-12
Update actions

turing85:


turing85:
Test was a success. I'll set this PR on draft. When publish-report in V2 is released, we can use this PR as base for the update since only the version for publish-report has to be changed from `@main` to `@v2`.

PR #432, opened 2024-03-08
[3.0.x] Update project version to 3.0.x-999-SNAPSHOT

zhfeng:


turing85 (approved):

turing85 (commented):
@zhfeng I don't think we can re-configure this now. We would need to do this with a release. The pipeline assumes that when `.github/project.yml` is changed, a release should be built. A pre-release pipeline ran on this PR.

Are you ok if we close this PR and do this change with the next release?

PR #461, opened 2024-03-25
Fix dependabot config

turing85:


zhfeng (approved):
It's no needs to backport to other branch. I think `dependabot` alway read the config from `main` branch.

turing85:
> It's no needs to backport to other branch. I think `dependabot` alway read the config from `main` branch.

Yes, it is true that it is not strictly needed to backport the change. It is more of a consistency-thing to have all branches reflect the same state to avoid confusion.

PR #560, opened 2024-07-11
Maintenance

turing85:
- Change deprecated dockerStatusBuildItem.isDockerAvailable() to dockerStatusBuildItem.isContainerRuntimeAvailable()
- Resource Adapter integration tests:
  - Where possible, use constructor injection
  - Remove public modifier from all test classes and -method
  - Reformat code
- Fix typos


turing85:
> I am missing this one:
> 
> > Change DockerStatusBuildItem to ContainerRuntimeStatusBuildItem

Yeah, sorry about that. It was initial part of the commit, but was removed, since `ContainerRuntimeStatusBuildItem` is abstract and cannot be used on its own. I updated the PR description and the commit message.

PR #624, opened 2024-09-09
#623: Restructure pipeline so PR runs from forks succeed.

### LINKED ISSUE #623: Workflow: restructure workflow so it does not fail on pull requests from forks
Currently, when we receive a pull request from a fork, the workflow will fail since the commenting on PRs does not work. The action [`publish-report` (`github.com`)](https://github.com/turing85/publish-report) was recently updated to allow this feature. 

We want to restructure our PR so it can work on PRs from forks with the new version of `publish-report`.

turing85:
Resolves #623.

Please do not be surprised. For the changes to take full effect, the workflows `pr-comment.yml` and `pr-report.yml` need to be on the default branch.

The final result will be virtually indistinguishable from the current behaviour, see https://github.com/turing85/quarkus-artemis/pull/2#issuecomment-2339250938

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,21 +24,27 @@ on:
       - '*.txt'
       - '.all-contributorsrc'
 
-permissions:
-  actions: write
-  checks: write
-  pull-requests: write
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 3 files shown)


zhfeng (approved):
Nice work!

gastaldi (approved):
Cool!

PR #679, opened 2024-12-19
Use and document Artemis BOM, fixes #437

### LINKED ISSUE #437: Support simple way to override/define the client version
As you mentioned in #402, we can/should define the Artemis client version ourselves - normally by using the same version as for the broker.

It would be good to have a simple way to set the client version instead of overriding the dependency management (with all the exclusions) as described in [this comment](https://github.com/quarkiverse/quarkus-artemis/issues/402#issuecomment-1938274938). For example: I don't know what exclusions might be required for an older or future client version.

It would be nice so have the possibility to set for example a Maven property which defines the client version. Also the import of a POM definition, which has the correct configuration per client could be a possibility.

middagj:
Also changed the `quarkus.version` into `quarkus-version` to align with Quarkus itself. And fix the `$` of `quarkus-artemis-version`.

Resolves #437

### DIFF
--- a/docs/modules/ROOT/pages/quarkus-artemis-ra.adoc
+++ b/docs/modules/ROOT/pages/quarkus-artemis-ra.adoc
@@ -16,16 +16,23 @@ it is recommended to add the following BOM to your project, *below the Quarkus B
 <dependencyManagement>
     <dependencies>
         <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-bom</artifactId>
+            <version>${artemis-version}</version>
+            <type>pom</type>
+            <scope>import</scope>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus.platform</groupId>
             <artifactId>quarkus-bom</artifactId>
-            <version>${quarkus.version}</version>
+            <version>${quarkus-version}</version>
             <type>pom</type>
             <scope>import</scope>
         </dependency>
         <dependency>
             <groupId>io.quarkiverse.artemis</groupId>
             <artifactId>quarkus-artemis-bom</artifactId>
-            <version>{quarkus-artemis-version}</version>
+            <version>${quarkus-artemis-version}</version>
             <type>pom</type>
             <scope>import</scope>
         </dependency>
@@ -43,7 +50,9 @@ Once the BOM is added, simply add the `io.quarkiverse.artemis:quarkus-artemis-ra
 </dependency>
 ----
 
-We recommend to set properties `quarkus.version` and `artemis.version` to the versions you are using/need to use. We also recommend to align `artemis.version` with the artemis server version used.
+We recommend to set properties `quarkus.version` and `artemis.version` to the versions you are using/need to use.
+We also recommend to align `artemis.version` with the artemis server version used.
+
 
 == Configuration
 
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -52,6 +52,14 @@
 
             <dependency>
                 <groupId>org.apache.activemq</groupId>
+                <artifactId>artemis-bom</artifactId>
+                <version>${artemis.version}</version>
+                <type>pom</type>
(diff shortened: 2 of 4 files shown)


docs/modules/ROOT/pages/quarkus-artemis-ra.adoc
```diff
@@ -43,7 +50,9 @@ Once the BOM is added, simply add the `io.quarkiverse.artemis:quarkus-artemis-ra
 </dependency>
 ----
 
-We recommend to set properties `quarkus.version` and `artemis.version` to the versions you are using/need to use. We also recommend to align `artemis.version` with the artemis server version used.
+We recommend to set properties `quarkus.version` and `artemis.version` to the versions you are using/need to use.
```
turing85, line 53: Does this actually work? Have we tested this?

middagj: Yes, I tested this with the integration test. Was planing on to do this with an example project as well.

turing85: Okay. I'd say let's see what your example project says.

middagj: Artemis bom import needs to precede the Quarkus one, but other than that it works. I also checked native image. This is the dependency:tree diff of an example file:

```diff
@@ -106,8 +106,8 @@
    +- io.quarkiverse.artemis:quarkus-artemis-core:jar:999-SNAPSHOT:compile
    |  +- io.quarkus:quarkus-netty:jar:3.17.5:compile
    |  |  \- io.netty:netty-codec:jar:4.1.115.Final:compile
-   |  \- org.apache.activemq:artemis-core-client:jar:2.38.0:compile
-   |     +- org.jgroups:jgroups:jar:5.3.13.Final:compile
+   |  \- org.apache.activemq:artemis-core-client:jar:2.37.0:compile
+   |     +- org.jgroups:jgroups:jar:5.3.10.Final:compile
    |     +- io.netty:netty-transport-native-epoll:jar:linux-x86_64:4.1.115.Final:compile
    |     +- io.netty:netty-transport-classes-epoll:jar:4.1.115.Final:compile
    |     +- io.netty:netty-transport-native-kqueue:jar:osx-x86_64:4.1.115.Final:compile
@@ -117,9 +117,9 @@
    |        +- org.bouncycastle:bcprov-jdk18on:jar:1.79:compile
    |        \- org.bouncycastle:bcpkix-jdk18on:jar:1.79:compile
    |           \- org.bouncycastle:bcutil-jdk18on:jar:1.79:compile
-   \- org.apache.activemq:artemis-jakarta-client:jar:2.38.0:compile
-      +- org.apache.activemq:artemis-commons:jar:2.38.0:compile
+   \- org.apache.activemq:artemis-jakarta-client:jar:2.37.0:compile
+      +- org.apache.activemq:artemis-commons:jar:2.37.0:compile
       |  \- commons-beanutils:commons-beanutils:jar:1.9.4:compile
       |     \- commons-collections:commons-collections:jar:3.2.2:compile
-      +- org.apache.activemq:artemis-selector:jar:2.38.0:compile
+      +- org.apache.activemq:artemis-selector:jar:2.37.0:compile
       \- jakarta.jms:jakarta.jms-api:jar:3.1.0:compile
```

and the native image produces `2.37.0` via `VersionLoader.getVersion().getFullVersion()`

turing85 (approved):

middagj:
How does it work, cherry pick the commit on the branches?

turing85:
Yes, and then push (without a PR).
I just checked, we cannot downstream to `3.2.x` since it still uses artemi `2.32.0` and the bom is only available starting with `2.33.0`. So we only need to downstream to `3.5.x`.

middagj:
I see you were ahead of me thanks!

turing85:
np :wink:

PR #741, opened 2025-02-18
Add support for exposing the Artemis Web-Ui

YuuichiMizuoka:
### Context
While using this Project and its provided Dev-Services I noticed that the Artemis web-ui (8161) is not exposed.
Since this is a feature I would find convenient I decided to build a version that would support it 

Other Dev-Services (Quarkus AMQP) do expose the Web-Ui, I was therefore curios if maybe this Project also has an Interested in adding the feature into its code base as well.

### Description
This PR now exposes the Artemis-Web-Ui (Port 8161) per default at a random port. 
Using the optional "web-ui-port" property a fixed port on which to expose the Ui can be defined.

### Changes in the Codebase
- Add the Optional Parameter `webUiPort` into the `ArtemisDevServicesBuildTimeConfig`
- Add `ARTEMIS_WEB_UI_PORT` to the exposed ports when building the `ArtemisContainer`
- Add the fixed exposed port (if set) when the  configuration is called

### If there is an Interested in this, i would like to kindly ask:
- if you could nudge me in the right direction towards how you would like to test this feature. (A quick "web-ui" can be reached via http request? Or something more in depth?)
- if you could nudge me towards how the documentation can be updated (I noticed the adoc files seem to usually just updated with the ci pipeline?)

### Considerations
This PR tries to keep the behavior of the `web-ui-port` property the same as `port`. Maybe it could also make sense to not expose 8161 per default (to keep the dev-services current behavior) and only expose them when explicitly set.

Thanks a lot for your time and for making this project!

PS: This is my second ever public PR, I kindly ask for feedback should there be any mistakes/breach of common protocol

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
@@ -1,10 +1,7 @@
 package io.quarkus.artemis.core.deployment;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 
@@ -43,8 +40,8 @@ import io.smallrye.config.NameIterator;
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public class DevServicesArtemisProcessor {
     private static final Logger LOGGER = Logger.getLogger(DevServicesArtemisProcessor.class);
-    private static final String QUARKUS_ARTEMIS_URL = "quarkus.artemis.url";
-    private static final String QUARKUS_ARTEMIS_NAMED_URL_TEMPLATE = "quarkus.artemis.%s.url";
+    private static final String QUARKUS_ARTEMIS_BASE = "quarkus.artemis.";
+    private static final String QUARKUS_ARTEMIS_NAMED_BASE_TEMPLATE = "quarkus.artemis.%s.";
 
     /**
      * Label to add to shared Dev Service for ActiveMQ Artemis running in containers.
@@ -181,10 +178,13 @@ public class DevServicesArtemisProcessor {
         cfgs.put(name, Objects.requireNonNull(configuration));
 
         if (devServices.get(name).isOwner()) {
-            LOGGER.infof(
-                    "Dev Services for ActiveMQ Artemis and named configuration %s started on %s",
-                    name,
-                    getArtemisUrl(name));
+            LOGGER.infof("Dev Services for ActiveMQ Artemis and named configuration %s started on %s",
+                    name, getArtemisUrl(name));
+
+            if (getArtemisWebUiUrl(name) != null) {
+                LOGGER.infof("Artemis Management Console for named configuration %s is available at %s",
+                        name, getArtemisWebUiUrl(name));
+            }
         }
         return devServices.get(name).toBuildItem();
     }
@@ -193,11 +193,23 @@ public class DevServicesArtemisProcessor {
         return devServices.get(name).getConfig().get(getUrlPropertyName(name));
     }
 
+    private static String getArtemisWebUiUrl(String name) {
+        return devServices.get(name).getConfig().get(getWebUiUrlPropertyName(name));
+    }
+
     private static String getUrlPropertyName(String name) {
+        return getArtemisPropertyBase(name) + "url";
+    }
+
+    private static String getWebUiUrlPropertyName(String name) {
+        return getArtemisPropertyBase(name) + "web-ui-url";
+    }
+
+    private static String getArtemisPropertyBase(String name) {
         if (Objects.equals(ArtemisUtil.DEFAULT_CONFIG_NAME, name)) {
-            return QUARKUS_ARTEMIS_URL;
+            return QUARKUS_ARTEMIS_BASE;
         } else {
-            return String.format(QUARKUS_ARTEMIS_NAMED_URL_TEMPLATE, name);
+            return String.format(QUARKUS_ARTEMIS_NAMED_BASE_TEMPLATE, name);
         }
     }
 
@@ -270,8 +282,10 @@ public class DevServicesArtemisProcessor {
                     containerName,
                     container.getContainerId(),
                     container::close,
-                    urlPropertyName,
-                    String.format("tcp://%s:%d", container.getHost(), container.getPort()));
+                    Map.of(
+                            urlPropertyName, String.format("tcp://%s:%d", container.getHost(), container.getPort()),
+                            getWebUiUrlPropertyName(name), String.format("http://%s:%d", container.getHost(),
+                                    container.getMappedPort(ARTEMIS_WEB_UI_PORT))));
         };
 
         return maybeContainerAddress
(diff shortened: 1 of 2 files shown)


turing85:
Hi @YuuichiMizuoka !
I really like the idea of exposing the web UI. However, there are two things I think we should address
- First, exposing the web UI as-is will not help since jolokia blocks the access. We would need to add `--relax-jolokia` to the extra args. We could add it to the default value [here (`github.com`)](https://github.com/quarkiverse/quarkus-artemis/blob/main/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java#L100). This, however, would run the risk that if the user overrides the extra args, the web UI stops working b/c they did not add `--relax-jolokia`. Another possibility would be to add it to `extraArgs` if it is not yet present [here (`github.com`)](https://github.com/quarkiverse/quarkus-artemis/blob/main/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java#L295). @middagj I'd like to have your opinion on this one.
- Second, a small log message a'la `Artemis web UI is available at http://localhost:<port-goes-here>/console/auth/login` would be nice so we do not have to look up the port (if it is randomly assigned). I think the log-message should be generatedt between [those two lines (`github.com`)](https://github.com/quarkiverse/quarkus-artemis/blob/main/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java#L250-L251).

middagj:
I would go for adding `--relax-jolokia` to extra args if not already present. It is the most dev friendly way. We could even do that for the other default args as well. I would even consider splitting up our `extraArgs` into two or more to remove the knowledge of the default. If one wants to add an additional argument, they probably won't consider the default.

Something like

`defaultExtraArgs` (`--no-autotune --mapped --no-fsync --relax-jolokia`)
`additionalExtraArgs` (``)

If a user overrides the `defaultExtraArgs` they are on their own.

YuuichiMizuoka:
Thanks for the fast feedback n.n
I will add the changes and update the PR when they are ready.

@middagj regarding the `additionalExtraArgs`:  would it make sense to just use the current `extraArgs` as the `additionalExtraArgs` parameter to avoid any breaking changes in the config?  

Alternatively, all 3 parameters could also be supported. with the `extraArgs` always just overwriting both new params, this would make sure all existing setups are for sure still going to work. (e.g if they set the extra args explicitly to remove a current default value)

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisDevServicesBuildTimeConfig.java
```diff
@@ -76,6 +83,10 @@ default int getPort() {
         return port().orElse(0);
     }
 
+    default int getWebUiPort() {
+        return webUiPort().orElse(0);
```
zhfeng, line 87: Why the default port is `0` with randomly chosen but not `8161`?

YuuichiMizuoka: The idea was to avoid port collisions on already existing CI systems who might already be running multiple pods in parallel. I agree that having 8161 as a default would be a nicer experience for local development.

zhfeng: OK, it makes sense.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -41,15 +38,16 @@
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 public class DevServicesArtemisProcessor {
     private static final Logger LOGGER = Logger.getLogger(DevServicesArtemisProcessor.class);
-    private static final String QUARKUS_ARTEMIS_URL = "quarkus.artemis.url";
-    private static final String QUARKUS_ARTEMIS_NAMED_URL_TEMPLATE = "quarkus.artemis.\"%s\".url";
+    private static final String QUARKUS_ARTEMIS_BASE = "quarkus.artemis.";
+    private static final String QUARKUS_ARTEMIS_NAMED_BASE_TEMPLATE = "quarkus.artemis.\"%s\".";
```
zhfeng, line 42: This could be fixed in https://github.com/quarkiverse/quarkus-artemis/pull/744. So please make sure to rebase your PR with the main branch if it includes the fix.

YuuichiMizuoka:
Hi there! I have updated the PR based on your feedback.

@turing85 the suggested place for the log did work, but it was only visible for a short amount of time before it would be rolled up together with the other logs. I could not find a simple way to exclude a certain log from this behavior.

So instead the log was moved to [here](https://github.com/quarkiverse/quarkus-artemis/blob/169b576bace496a8387ed5517b949d3d98afc6d4/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java#L169), together with the other url log. (Using the config map the hold the url as a return value)


@middagj How far would you like to go regarding the matcher? I pushed a very naive parsing, that seems to work well for most use cases, breaks apart though if:
  - The user uses key-value arguments in the `defaultExtraArgs` that they duplicate into the `extraArgs` as well
  - The user has a key-value argument in the extraArgs that contain the name of one of our default args as value (e.g. `extra-args=--cluster-user="--relax-jolokia"`)

But both of those seem quite unlikely, and fixable by the user, so it might be a worthwhile trade off for code simplicity? 
(Also, in any case, its an improvement over the current handling where all the defaults are just gone)


Please let me know your thought on the update, happy to add more changes if needed!

middagj:
> @middagj How far would you like to go regarding the matcher? I pushed a very naive parsing, that seems to work well for most use cases, breaks apart though if:
> 
>     * The user uses key-value arguments in the `defaultExtraArgs` that they duplicate into the `extraArgs` as well
> 
>     * The user has a key-value argument in the extraArgs that contain the name of one of our default args as value (e.g. `extra-args=--cluster-user="--relax-jolokia"`)
> 
> 
> But both of those seem quite unlikely, and fixable by the user, so it might be a worthwhile trade off for code simplicity? (Also, in any case, its an improvement over the current handling where all the defaults are just gone)
> 
> Please let me know your thought on the update, happy to add more changes if needed!

I would go indeed for simple parsing. The current looks good. Perhaps issue a warning when we filter. You could also argue that because we have now split up, we can just add and issue warning if we find duplicates (if Artemis doesn't break). I was more iterating on ideas now composing them.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
YuuichiMizuoka: thanks for the heads-up, i rebased the branch n.n

YuuichiMizuoka:
@middagj unfortunately Artemis crashes when a duplicated param is given, so it seems like we will have to keep our filtering

I also like the log idea, I pushed a commit adding that n.n

turing85 (approved):

zhfeng (approved):
LGTM

middagj (approved):
🚀

turing85:
@YuuichiMizuoka I downstreamed the changes to branches `3.5.x` and `3.2.x`. I will produce new releases later today :tada: 

Thanks again for the contribution!

turing85:
@YuuichiMizuoka the improvement has been released with versions `3.2.7`, `3.5.6` and `3.7.1` :tada:

YuuichiMizuoka:
> @YuuichiMizuoka I downstreamed the changes to branches `3.5.x` and `3.2.x`. I will produce new releases later today 🎉
> 
> Thanks again for the contribution!

Yay, thanks!

And thanks a lot to everyone for all the help and feedback! n.n

PR #744, opened 2025-02-20
Fix #735 to omit the quotes in artemis url generated by devservice if…

### LINKED ISSUE #735: test using dev service with named configuration fails with application.yaml
When running tests with devservice using a named configuration in an `application.yaml`, the url fails to be set.
When instead using an `application.properties` it succeeds.
Here is a reproducer: https://github.com/vsevel/reproducer_artemis_dev_service_config_yaml
cc @zhfeng

zhfeng:
… using yaml configuration

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
@@ -8,6 +8,7 @@ import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Supplier;
 
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.logging.Logger;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.Network;
@@ -34,6 +35,7 @@ import io.quarkus.devservices.common.ConfigureUtil;
 import io.quarkus.devservices.common.ContainerAddress;
 import io.quarkus.devservices.common.ContainerLocator;
 import io.quarkus.runtime.configuration.ConfigUtils;
+import io.smallrye.config.NameIterator;
 
 /**
  * Start a ActiveMQ Artemis broker if needed
@@ -42,7 +44,7 @@ import io.quarkus.runtime.configuration.ConfigUtils;
 public class DevServicesArtemisProcessor {
     private static final Logger LOGGER = Logger.getLogger(DevServicesArtemisProcessor.class);
     private static final String QUARKUS_ARTEMIS_URL = "quarkus.artemis.url";
-    private static final String QUARKUS_ARTEMIS_NAMED_URL_TEMPLATE = "quarkus.artemis.\"%s\".url";
+    private static final String QUARKUS_ARTEMIS_NAMED_URL_TEMPLATE = "quarkus.artemis.%s.url";
 
     /**
      * Label to add to shared Dev Service for ActiveMQ Artemis running in containers.
@@ -72,6 +74,20 @@ public class DevServicesArtemisProcessor {
             DevServicesConfig devServicesConfig) {
         ArrayList<DevServicesResultBuildItem> results = new ArrayList<>();
         for (String name : bootstrap.getConfigurationNames()) {
+            String propertyName = name;
+            for (String rawName : ConfigProvider.getConfig().getPropertyNames()) {
+                if (rawName.contains(name)) {
+                    rawName = rawName.substring("quarkus.artemis.".length());
+                    NameIterator nameIterator = new NameIterator(rawName);
+                    if (nameIterator.nextSegmentEquals(name)) {
+                        int end = nameIterator.getNextEnd();
+                        if (rawName.charAt(0) == '"' && rawName.charAt(end - 1) == '"') {
+                            propertyName = rawName.substring(0, end);
+                            break;
+                        }
+                    }
+                }
+            }
             ArtemisBuildTimeConfig buildTimeConfig = buildConfigs.configs().get(name);
             boolean isUrlEmpty = shadowRunTimeConfigs.isUrlEmpty(name);
             if (!shadowRunTimeConfigs.getNames().contains(name) && buildTimeConfig.isEmpty()) {
@@ -86,7 +102,7 @@ public class DevServicesArtemisProcessor {
                     isUrlEmpty);
             DevServicesResultBuildItem result = start(
                     configuration,
-                    name,
+                    propertyName,
                     dockerStatusBuildItem,
                     launchMode,
                     consoleInstalledBuildItem,
--- a/core/deployment/pom.xml
+++ b/core/deployment/pom.xml
@@ -49,6 +49,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.eclipse.microprofile.config</groupId>
+            <artifactId>microprofile-config-api</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>io.quarkiverse.artemis</groupId>
             <artifactId>quarkus-artemis-core</artifactId>
         </dependency>


radcortez (approved):

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -72,6 +73,17 @@ public List<DevServicesResultBuildItem> startArtemisDevService(
             DevServicesConfig devServicesConfig) {
         ArrayList<DevServicesResultBuildItem> results = new ArrayList<>();
         for (String name : bootstrap.getConfigurationNames()) {
+            String propertyName = name;
+            for (String rawName : ConfigProvider.getConfig().getPropertyNames()) {
+                if (rawName.contains(name)) {
+                    String str = rawName.substring("quarkus.artemis.".length());
+                    str = str.substring(0, str.indexOf('.'));
+                    if (str.replace("\"", "").replace("'", "").equals(name)) {
```
turing85, line 81: Should we use `String.replace(char, char)` here?

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -72,6 +73,17 @@ public List<DevServicesResultBuildItem> startArtemisDevService(
             DevServicesConfig devServicesConfig) {
         ArrayList<DevServicesResultBuildItem> results = new ArrayList<>();
         for (String name : bootstrap.getConfigurationNames()) {
+            String propertyName = name;
+            for (String rawName : ConfigProvider.getConfig().getPropertyNames()) {
+                if (rawName.contains(name)) {
+                    String str = rawName.substring("quarkus.artemis.".length());
+                    str = str.substring(0, str.indexOf('.'));
```
turing85, line 80: This is not going to work. In theory, a name like `quarkus.artemis."my.connection"...` is legal. If I understand the algorithm correctly, `str` would be `my`, while the `propertyName`  would be `my.connection`.

If we want to do it properly, I fear we have to do some parsing...

@radcortez WDYT?

turing85 (changes requested):

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
turing85: I spoke with @radcortez and he recommends using `io.smallrye.config.NameIterator` to analyze the config name.

zhfeng: Yeah, I will try to use `NameIterator`.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
zhfeng: No, we need to remove the quotes here.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
zhfeng: Also `"foo.'1'"` is valid property name?

radcortez: Yes, `NameIterator` removes quotes from the segments unless they are escaped.

turing85: @zhfeng the `NameIterator` should provide access to the offsets of a segment. So to figure out whether the name is quoted, we'd need to check the characters before and after the segment.

zhfeng: Thanks all - I update the algorithm to use `nameIterator`, please take a review again!

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
zhfeng: I think we don't need this anymore

vsevel:
I checked the current branch, but I am still seeing the issue.

```
git branch
* issue_735
```

```
$ ls -al quarkus-sample-camel/target/quarkus-app/lib/main/ | grep artemis
-rw-r--r-- 1 Sevel 1049089   22077 Feb 25 17:17 io.quarkiverse.artemis.quarkus-artemis-core-999-SNAPSHOT.jar
-rw-r--r-- 1 Sevel 1049089    8017 Feb 25 17:17 io.quarkiverse.artemis.quarkus-artemis-jms-999-SNAPSHOT.jar
-rw-r--r-- 1 Sevel 1049089  671888 Jan 20 10:28 org.apache.activemq.artemis-commons-2.39.0.jar
-rw-r--r-- 1 Sevel 1049089  842043 Jan 20 10:28 org.apache.activemq.artemis-core-client-2.39.0.jar
-rw-r--r-- 1 Sevel 1049089   21627 Jan 20 10:28 org.apache.activemq.artemis-hqclient-protocol-2.39.0.jar
-rw-r--r-- 1 Sevel 1049089  190146 Jan 20 10:28 org.apache.activemq.artemis-jakarta-client-2.39.0.jar
-rw-r--r-- 1 Sevel 1049089  113104 Jan 20 10:28 org.apache.activemq.artemis-selector-2.39.0.jar
```

in `src/main/ressources/application.yaml`:
```
quarkus:
  artemis:
    "cf-in":
      enabled: true
    "cf-out":
      enabled: true
```

in `application-dev.yaml` :
```
  artemis:
    "cf-in":
      url: "x"
      username: "x"
      password: "x"
    "cf-out":
      url: "x"
      username: "x"
      password: "x"
```

anything I am missing?
anything I can do to help pinpoint the issue?

zhfeng:
How did you run the application? and can you paste the error messages?

vsevel:
one thing I forgot to mention is that this fails when evaluating a property using another property.
the error message is 
```
Caused by: java.util.NoSuchElementException: SRCFG00011: Could not expand value quarkus.artemis."cf-in".url in property myapp.camel.jms.dynamic-cf-in.url
```

`application.yaml`:
```
quarkus:
  artemis:
    "cf-in":
      enabled: true
    "cf-out":
      enabled: true

myapp:
  camel:
    jms:
      dynamic-cf-in:
        url: ${quarkus.artemis."cf-in".url}
        username: ${quarkus.artemis."cf-in".username}
        password: ${quarkus.artemis."cf-in".password}
```

and `application-dev.yaml`
```
quarkus:
  artemis:
    "cf-in":
      url: "x"
      username: "x"
      password: "x"
    "cf-out":
      url: "x"
      username: "x"
      password: "x"
```

zhfeng:
Hmm, since yaml ignores the quote, can you try
```
quarkus:
  artemis:
    '"cf-in"':
      enabled: true
    '"cf-out"':
      enabled: true
```

vsevel:
I gave up property indirection, and used that code in my route builder:
```
    @Override
    public void configure() {
        registerDynamicCF(
                getProperty("quarkus.artemis.cf-in.url"),
                getOptionalProperty("quarkus.artemis.cf-in.username"),
                getOptionalProperty("quarkus.artemis.cf-in.password"),
                "dynamic-cf-in");
...

    private void registerDynamicCF(String uri, Optional<String> user, Optional<String> password, String camelRegistryName) {
        log.info("creating dynamic jms connection factory with uri {}", uri);
        var poolCF = new JmsPoolLocalTransactionConnectionFactory();
        if (user.isPresent()) {
            poolCF.setConnectionFactory(new ActiveMQQueueConnectionFactory(uri, user.get(), password.orElseThrow()));
        } else {
            poolCF.setConnectionFactory(new ActiveMQQueueConnectionFactory(uri));
        }

        InstanceHandle<TransactionManager> instance = Arc.container().instance(TransactionManager.class);
        if (instance.isAvailable()) {
            poolCF.setTransactionManager(instance.get());
        }
        getContext().getRegistry().bind(camelRegistryName, poolCF);
    }
```

and this is working.
I do not know if we want to tackle the indirection use case as well. not sure if this is something in control of the extension, or deeper in MP Config.

zhfeng:
No, the yaml ignores `"` in the property name. So the dev service will propagate the url with `quarkus.artemis.cf-in.url`. So I think you need to use the same in the `dynamic-cf-in` with `url: ${quarkus.artemis.cf-in.url}`

zhfeng:
IIRC, it could be related to the yaml parser? @radcortez

vsevel:
I checked also that if I drop `"` everywhere it works fine. what is the purpose again of using `"`?

zhfeng:
@turing85 can you take a review? is there anything else we need to concern? if not, I will merge it this week. Thanks!

turing85 (approved):

turing85:
LGTM.

@radcortez can you take a look as well?

vsevel:
many extensions must have the same issue. it is a generic one. not something specific to artemis.
I am wondering how much could be moved to smallrye config, and applied transparently, or at least through calling some utility function.

zhfeng:
I agree and can you open an issue on [smallrye-config](https://github.com/smallrye/smallrye-config/issues) ?

turing85:
While I agree on an UX-level, I wonder what we can do on a technical level to make the experience better. Opening an issue/discussion in [smallrye-config](https://github.com/smallrye/smallrye-config/issues) - as @zhfeng said - is most probably the best way to go.

vsevel:
yes @radcortez I am aware of your comments. I am not thinking about completely automatic.
I was just wondering if there was some utility functions you could provide in SR Config, which would be used in those extensions, to hide part of the complexity.
it would still be the responsibility of the extensions to call your utility function, but they would not have to recode it entirely.
the utility function would have some documentation on why and why to call it.
and when coding a new extension, we could see calls being made from other extensions, and trigger some hints that something needs to be done.
I am just afraid that we are fixing it for artemis, and one by one we will need to chase bugs in other extensions.
if you do not think that something is possible, let's leave it there then. 
thanks @zhfeng for looking into it.

radcortez:
We can certainly include some utilities for this case. I'll have a look.

PR #829, opened 2025-06-06
Support Compose DevService

vsevel:
This brings some support for compose dev service in artemis jms RA.
I left 3 `TODO` with open questions. I would like your advice @ozangunalp 
I used apicurio as an example to bring those modifications from this [commit](https://github.com/quarkusio/quarkus/pull/46848/files#diff-a95c1e07d280f4aa2a52da9dc054fd1941c07ec5020ba046b0c75052d1ba7314)
I have tested with:
```
name: my-app
services:
  artemis:
    image: quay.io/arkmq-org/activemq-artemis-broker:artemis.2.40.0
    ports:
      - "61616"
    environment:
      AMQ_USER: admin
      AMQ_PASSWORD: admin
      AMQ_EXTRA_ARGS: --no-autotune --mapped --no-fsync
    labels:
      io.quarkus.devservices.compose.wait_for.ports.disable: true
      io.quarkus.devservices.compose.wait_for.logs: .*AMQ241004.*
```
this works fine. if the compose container is started, it gets picked up. if not, a regular container will be launched.
see https://quarkus.io/guides/compose-dev-services

### DIFF
--- a/docs/modules/ROOT/pages/quarkus-artemis-ra.adoc
+++ b/docs/modules/ROOT/pages/quarkus-artemis-ra.adoc
@@ -123,6 +123,26 @@ public class MyResourceEndpoint implements MessageListener {
 <1> This MessageListener will be activated by the `other` resource adapter configured above.
 <2> This `ConnectionFactory` will be the one configured in the `main` resource adapter.
 
+[[Compose]]
+== Compose
+
+The Artemis RA Dev Services supports xref:https://quarkus.io/guides/compose-dev-services[Compose Dev Services].
+It relies on a `compose-devservices.yml`, such as:
+
+[source,yaml]
+----
+name: <application name>
+services:
+  artemis:
+    image: quay.io/arkmq-org/activemq-artemis-broker:artemis.2.40.0
+    ports:
+      - "61616"
+    environment:
+      AMQ_USER: admin
+      AMQ_PASSWORD: admin
+      AMQ_EXTRA_ARGS: --no-autotune --mapped --no-fsync
+----
+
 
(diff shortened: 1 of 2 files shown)


turing85:
@vsevel is the PR ready for (general) review, or do you want a review from @ozangunalp first, possibly fix things and then give it to us (the quarkus-artemis team) to review?

vsevel:
I left 3 `TODO` in the code. a review from anybody with enough knowledge to tell me what to do, and how to test it would be nice. 
I already tested on my project. so I know it works at least in one situation.
the other solution is to remove the `TODO`, and to follow up PRs when somebody comes with an issue.

ozangunalp (changes requested):
Looks good overall. I prefer the optional chaining as it is mostly the pattern used in quarkus core.

I think the image name partial can be `artemis` or `activemq-artemis`.

vsevel:
thanks for the review @ozangunalp 
I took into account the different suggestions
I added some doc (similar to what we have in core)
I tested with dev mode in these situations: 
- no compose file
- compose file, but not started with `up`
- compose file started with `up` 
additionally I ran the tests with: 
- `mvn test` with and without compose file
- `mvn test -Dquarkus.compose.devservices.reuse-project-for-tests=true` with `up`
I believe we are reafy for the final review @turing85

ozangunalp (approved):

turing85:
@middagj, @zhfeng lgty?

middagj (approved):
LGTM

turing85:
@vsevel release is not yet available. Pipeline is still running.

PR #864, opened 2025-07-18
Do not use Runtime configuration during deployment

radcortez:
With https://github.com/quarkusio/quarkus/pull/48500, we will disallow the injection of Runtime configuration objects during deployment. The recommended way is to inject it directly into the Recorder wrapped into a `RuntimeValue`.


turing85 (approved):

gsmet:
@turing85 we would appreciate a new micro as we are relying on your extension for testing some Quarkus features. Thanks!

PR #953, opened 2025-11-05
Artemis test resource allocate dynamic port

alex-kovalenko1982:
A couple of improvements based on usage of an embedded artemis broker via ArtemisTestResource in real projects.

#### Support dynamic port allocation
It's not a rare case when the local development machine has already some processes bound to the default artemis port 61616. Including some left-overs or caches from previous test run. ActiveMQ supports dynamic port allocation when 0 value provided as an acceptor port in the broker.xml. This PR extends ArtemisTestResource to return the actual allocated port into the `quarkus.artemis.url` property.

```
<connectors>
	<connector name="activemq">tcp://localhost:0</connector>
</connectors>
<acceptors>
	<acceptor name="activemq">tcp://localhost:0</acceptor>
</acceptors>
```		

#### Simplified named test resource instantiation
Both `@QuarkusTestResource` and `@WithTestResource` annotations support initial arguments to instantiate a test resource. Thus, instead of created an extended class for a NamedArtemisTestResource the init argument with a configuration name can be provided together with a default ArtemisTestResource implementation.

```
@QuarkusTest
@WithTestResource(ArtemisTestResource.class)
@WithTestResource(value = ArtemisTestResource.class, initArgs = { @ResourceArg(name = "configurationName", value = "named-1") })
class EmbeddedHealthCheckTest extends BaseArtemisHealthCheckTest {
}
```

### DIFF
--- a/integration-tests/core/with-default/src/test/java/io/quarkus/it/artemis/core/withdefault/embedded/EmbeddedHealthCheckTest.java
+++ b/integration-tests/core/with-default/src/test/java/io/quarkus/it/artemis/core/withdefault/embedded/EmbeddedHealthCheckTest.java
@@ -1,8 +1,8 @@
 package io.quarkus.it.artemis.core.withdefault.embedded;
 
 import io.quarkus.artemis.test.ArtemisTestResource;
-import io.quarkus   .it.artemis.core.withdefault.BaseArtemisHealthCheckTest;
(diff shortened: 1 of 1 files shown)


zhfeng:
@alex-kovalenko1982 is it possible to add a test for the dynamic port?

turing85:
LGTM - thanks for the contibution @alex-kovalenko1982

PR #961, opened 2025-11-08
Move constant to separate class ArtemisConstants

turing85:


turing85:
> I am okay with the change, but have a personal preference to keep constants close to their use. Feel free to merge.

The reason I moved them is that I am using at least one of the constants in customer projects, and I forgot for a while that it existed (and used a magic constant instead). So these constants are (among other things) meant to be used by users of the extension. Having them in a central place makes it is - hopefully - easier to find them.

middagj:
> > I am okay with the change, but have a personal preference to keep constants close to their use. Feel free to merge.
> 
> The reason I moved them is that I am using at least one of the constants in customer projects, and I forgot for a while that it existed (and used a magic constant instead). So these constants are (among other things) meant to be used by users of the extension. Having them in a central place makes it is - hopefully - easier to find them.

Makes sense.

PR #1019, opened 2026-01-28
Added all ServiceLocator and ConnectionFactory settings as configs

mikethecalamity:
Now you can apply all the configs for `ServiceLocator` and `ConnectionFactory` via application properties.

For example:
```properties
quarkus.artemis.consumer-window-size=2048
quarkus.artemis.call-timeout=5000
quarkus.artemis.auto-group=true

quarkus.artemis."named-1".producer-max-rate=100
quarkus.artemis."named-1".retry-interval-multiplier=2.5
quarkus.artemis."named-1".pre-acknowledge=true
```

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisCoreRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisCoreRecorder.java
@@ -23,14 +23,65 @@ public class ArtemisCoreRecorder {
         ArtemisRuntimeConfig runtimeConfig = runtimeConfigs.getValue().configs().get(name);
         ArtemisBuildTimeConfig buildTimeConfig = buildTimeConfigs.configs().get(name);
         ArtemisUtil.validateIntegrity(name, runtimeConfig, buildTimeConfig);
-        ServerLocator serverLocator = Objects.requireNonNull(getServerLocator(runtimeConfig.getUrl()));
+        ServerLocator serverLocator = Objects.requireNonNull(getServerLocator(runtimeConfig));
         return () -> serverLocator;
     }
 
-    protected static ServerLocator getServerLocator(String url) {
-        if (url != null) {
+    protected static ServerLocator getServerLocator(ArtemisRuntimeConfig runtimeConfig) {
+        if (runtimeConfig.url().isPresent()) {
             try {
-                return ActiveMQClient.createServerLocator(url);
+                ServerLocator locator = ActiveMQClient.createServerLocator(runtimeConfig.url().get());
+
+                // --- Flow Control ---
+                runtimeConfig.consumerWindowSize().ifPresent(locator::setConsumerWindowSize);
+                runtimeConfig.consumerMaxRate().ifPresent(locator::setConsumerMaxRate);
+                runtimeConfig.producerWindowSize().ifPresent(locator::setProducerWindowSize);
+                runtimeConfig.producerMaxRate().ifPresent(locator::setProducerMaxRate);
+                runtimeConfig.confirmationWindowSize().ifPresent(locator::setConfirmationWindowSize);
+                runtimeConfig.minLargeMessageSize().ifPresent(locator::setMinLargeMessageSize);
+                runtimeConfig.compressLargeMessage().ifPresent(locator::setCompressLargeMessage);
+                runtimeConfig.compressionLevel().ifPresent(locator::setCompressionLevel);
+                runtimeConfig.cacheLargeMessagesClient().ifPresent(locator::setCacheLargeMessagesClient);
+
+                // --- Connectivity & Timeouts ---
+                runtimeConfig.clientFailureCheckPeriod().ifPresent(locator::setClientFailureCheckPeriod);
+                runtimeConfig.connectionTtl().ifPresent(locator::setConnectionTTL);
+                runtimeConfig.callTimeout().ifPresent(locator::setCallTimeout);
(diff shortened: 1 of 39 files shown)


middagj (commented):
Looks fine code wise. Some questions / minor things commented. If I understand correctly, the options are additionally to the url set.

The integration tests are not really my thing. I don't really understand why there need to be two. I think it would be best if someone else also takes a look.

integration-tests/core/with-locator-customization/src/main/java/io/quarkus/it/artemis/core/withlocatorcustomization/ArtemisEndpoint.java
```diff
@@ -0,0 +1,60 @@
...
+
+@Path("/artemis")
+@Produces(MediaType.TEXT_PLAIN)
+public class ArtemisEndpoint {
+    private final ServerLocator defaultLocator;
+    private final ServerLocator namedOneLocator;
+
+    public ArtemisEndpoint(
+            ServerLocator defaultLocator,
+            @Identifier("named-1") ServerLocator namedOneLocator) {
```
turing85: Please add `@SuppressWarnings("CdiInjectionPointsInspection")` to both parameters.

integration-tests/jms/with-factory-customization/src/main/java/io/quarkus/it/artemis/core/withfactorycustomization/ArtemisEndpoint.java
```diff
@@ -0,0 +1,62 @@
...
+
+@Path("/artemis")
+@Produces(MediaType.TEXT_PLAIN)
+public class ArtemisEndpoint {
+    private final ActiveMQConnectionFactory defaultFactory;
+    private final ActiveMQConnectionFactory namedOneFactory;
+
+    public ArtemisEndpoint(
+            ConnectionFactory defaultFactory,
+            @Identifier("named-1") ConnectionFactory namedOneFactory) {
```
turing85: Please add `@SuppressWarnings("CdiInjectionPointsInspection")` to both parameters.

turing85 (approved):
Integration tests LGTM. There are minor things (see below), but no show stoppers.

turing85:
> Looks fine code wise. Some questions / minor things commented. If I understand correctly, the options are additionally to the url set.
> 
> The integration tests are not really my thing. I don't really understand why there need to be two. I think it would be best if someone else also takes a look.

@middagj What exactly do you mean by "two" integration tests? Do you mean the part in core and the part in jms? Or do you mean why two connection factories are used?

middagj:
> @middagj What exactly do you mean by "two" integration tests? Do you mean the part in core and the part in jms? Or do you mean why two connection factories are used?

I mean the `with-locator-customization` and the `with-factory-customization`. I missed the part that it was one for core and one for JMS. I resolved the comment about it as well.

core/runtime/src/main/java/io/quarkus/artemis/core/runtime/ArtemisCoreRecorder.java
```diff
@@ -23,14 +23,65 @@ public Supplier<ServerLocator> getServerLocatorSupplier(String name) {
...
         ArtemisUtil.validateIntegrity(name, runtimeConfig, buildTimeConfig);
-        ServerLocator serverLocator = Objects.requireNonNull(getServerLocator(runtimeConfig.getUrl()));
+        ServerLocator serverLocator = Objects.requireNonNull(getServerLocator(runtimeConfig));
         return () -> serverLocator;
     }
 
-    protected static ServerLocator getServerLocator(String url) {
-        if (url != null) {
+    protected static ServerLocator getServerLocator(ArtemisRuntimeConfig runtimeConfig) {
+        if (runtimeConfig.url().isPresent()) {
```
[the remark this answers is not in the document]
mikethecalamity: I do agree that is possible, but in this case it doesn't change much and is kind of awkward because the try-catch needs to live inside the `map()`.

here is an example of what that would look like:
```java
ServerLocator locator = runtimeConfig.url().map(u -> {
    try {
        return ActiveMQClient.createServerLocator(u);
    } catch (Exception e) {
        throw new RuntimeException(e);
    }
}).orElse(null);

if (locator!= null) {
    //setters
}

return locator;
```

But if you do prefer this, I can change it.

integration-tests/jms/with-factory-customization/src/main/java/io/quarkus/it/artemis/core/withfactorycustomization/ArtemisEndpoint.java
mikethecalamity: done

integration-tests/core/with-locator-customization/src/main/java/io/quarkus/it/artemis/core/withlocatorcustomization/ArtemisEndpoint.java
mikethecalamity: done

turing85:
@mikethecalamity FYI: Release `3.12.1` is on its way. It contains a bugfix (see #1027). I took the liberty and renamed properties `deserializationWhiteList()` and `deserializationBackList()` to `deserializationAllowList()` and `deserializationDenyList()` (See commit https://github.com/quarkiverse/quarkus-artemis/commit/d54a1974081dd27aea62a57642d4bc1d1dfed4fa)

PR #1047, opened 2026-02-19
Migrate RA Dev Service to the new API

### LINKED ISSUE #874: Replace deprecated `RunningDevService`
As far as I can see, we use `RunningDevService` in multiple places in two places:
- https://github.com/quarkiverse/quarkus-artemis/blob/a5b5942f32194b532c08771972935c38e7d55fae/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java, and
- https://github.com/quarkiverse/quarkus-artemis/blob/a5b5942f32194b532c08771972935c38e7d55fae/ra/deployment/src/main/java/io/quarkus/artemis/jms/ra/deployment/DevServicesArtemisProcessor.java

`RunningDevService` was deprecated, thus we should replace it.

Some information can be found [here (`quarkusio.zulichat.com`)](https://quarkusio.zulipchat.com/#narrow/channel/187030-users/topic/Deprecation.20of.20.60RunningDevService.60/with/529525657) and [here (`quarkusio.zulipchat.com`)](https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/New.20DevServicesResultBuildItem.20owned.2Fdiscovered.20API/with/528814001).

vsevel:
this is an adaptation to the new dev service api.
I used claude for the initial draft, then reviewed and adapted it.
then I tested it against one of our application in 3 different configurations:
- normal (ie : not compose)
- with a compose file, but letting quarkus launch the container
- with a compose file where I do a compose up prior to launching the test, and using property quarkus.compose.devservices.reuse-project-for-tests=true (and again with value false)

I used https://github.com/quarkusio/quarkus/blob/main/extensions/redis-client/deployment/src/main/java/io/quarkus/redis/deployment/client/DevServicesRedisProcessor.java as an inspiration
here is my compose file:
```
name: jms-new-dev-service
services:
  artemis:
    image: [quay.io/arkmq-org/activemq-artemis-broker:artemis.2.40.0](http://quay.io/arkmq-org/activemq-artemis-broker:artemis.2.40.0)
    ports:
    - '61616'
    environment:
      AMQ_USER: admin
      AMQ_PASSWORD: admin
      AMQ_EXTRA_ARGS: --no-autotune --mapped --no-fsync
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,12 +110,6 @@ jobs:
             --batch-mode \
             --define maven.buildNumber.skip \
             de.qaware.maven:go-offline-maven-plugin:resolve-dependencies
-          ./mvnw \
-            --batch-mode \
-            --define maven.buildNumber.skip \
-            --projects integration-tests \
-            --also-make-dependents \
-            quarkus:go-offline
 
   build-and-test-jvm:
     runs-on: ubuntu-latest
--- a/docs/modules/ROOT/pages/includes/attributes.adoc
+++ b/docs/modules/ROOT/pages/includes/attributes.adoc
@@ -1,4 +1,4 @@
-:quarkus-version: 3.34.0
+:quarkus-version: 3.34.3
 :quarkus-artemis-version: 3.14.1
 
 :quarkus-org-url: https://github.com/quarkusio
--- a/docs/modules/ROOT/pages/includes/quarkus-artemis-jms-ra.adoc
+++ b/docs/modules/ROOT/pages/includes/quarkus-artemis-jms-ra.adoc
@@ -1,0 +1,236 @@
+[.configuration-legend]
+icon:lock[title=Fixed at build time] Configuration property fixed at build time - All other configuration properties are overridable at runtime
+[.configuration-reference.searchable, cols="80,.^10,.^10"]
+|===
+
+h|[.header-title]##Configuration property##
+h|Type
+h|Default
+
+a|icon:lock[title=Fixed at build time] [[quarkus-artemis-jms-ra_quarkus-ironjacamar-devservices-enabled]] [.property-path]##link:#quarkus-artemis-jms-ra_quarkus-ironjacamar-devservices-enabled[`+++quarkus.ironjacamar.devservices.enabled+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.ironjacamar.devservices.enabled+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Enable or disable Dev Services explicitly. Dev Services are automatically enabled unless `quarkus.ironjacamar.devservices.enabled` is set.
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_IRONJACAMAR_DEVSERVICES_ENABLED+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_IRONJACAMAR_DEVSERVICES_ENABLED+++`
+endif::add-copy-button-to-env-var[]
+--
+|boolean
+|
+
+a|icon:lock[title=Fixed at build time] [[quarkus-artemis-jms-ra_quarkus-ironjacamar-devservices-reuse]] [.property-path]##link:#quarkus-artemis-jms-ra_quarkus-ironjacamar-devservices-reuse[`+++quarkus.ironjacamar.devservices.reuse+++`]##
+ifdef::add-copy-button-to-config-props[]
+config_property_copy_button:+++quarkus.ironjacamar.devservices.reuse+++[]
+endif::add-copy-button-to-config-props[]
+
+
+[.description]
+--
+Allow to reuse Artemis container between dev mode sessions.
+
+
+ifdef::add-copy-button-to-env-var[]
+Environment variable: env_var_with_copy_button:+++QUARKUS_IRONJACAMAR_DEVSERVICES_REUSE+++[]
+endif::add-copy-button-to-env-var[]
+ifndef::add-copy-button-to-env-var[]
+Environment variable: `+++QUARKUS_IRONJACAMAR_DEVSERVICES_REUSE+++`
+endif::add-copy-button-to-env-var[]
+--
+|boolean
+|`+++false+++`
+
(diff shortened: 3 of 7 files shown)


turing85:
So funny thing... running `./mvn -DskipTests clean package` works just fine, while `./mvnw quarkus:go-offline` results in the above error.

@holly-cummins could it be that something is broken in `quarkus-devservices` with respect to quarkus' `go-offline` target?

vsevel:
I created https://github.com/quarkusio/quarkus/issues/52695

turing85:
@holly-cummins it seems that your ~~workaroun~~ fix does not work...

vsevel:
the full build is successful on my host.

turing85:
> the full build is successful on my host.

Try running

```bash
./mvnw \
  --batch-mode \
  --define maven.buildNumber.skip \
  --projects integration-tests \
  --also-make-dependents \
  quarkus:go-offline
```

This will trigger the error.

vsevel:
I don't doubt it. I was just saying that I checked all tests were running fine. 
I did not build an app against it yet to make sure sure it is working fine (beside the IT of the project).
I will do this next week.
there is deprecated warning I want to improve also.

holly-cummins:
What's annoying is that the workaround wasn't clever, it was crude ("don't call the thing that doesn't work"), and yet somehow it still ended up calling the thing. Doh! 

Looking at those logs, the build is still calling `go-offline` on all projects, and I'm unsure why. :(

turing85:
(forget what I wrote; I tested on the wrong branch)

vsevel:
> What's annoying is that the workaround wasn't clever, it was crude ("don't call the thing that doesn't work"), and yet somehow it still ended up calling the thing. Doh!
> 
> Looking at those logs, the build is still calling `go-offline` on all projects, and I'm unsure why. :(

isn't this simply related to https://github.com/quarkiverse/quarkus-artemis/blob/main/.github/workflows/build.yml#L118?

holly-cummins:
> > What's annoying is that the workaround wasn't clever, it was crude ("don't call the thing that doesn't work"), and yet somehow it still ended up calling the thing. Doh!
> > Looking at those logs, the build is still calling `go-offline` on all projects, and I'm unsure why. :(
> 
> isn't this simply related to https://github.com/quarkiverse/quarkus-artemis/blob/main/.github/workflows/build.yml#L118?

Sort of. That's definitely the offending line, but I don't entirely know why it's offending. :) 

What I *thought* the workaround was doing was changing the build scripts to only run `go-offline` for the integration test projects, not the main extension projects, where it doesn't make sense/isn't supported. And looking at the build logs, that seems to be what's happening. But things are still failing. I haven't read the logs in enough detail to figure out which of my assumptions is wrong.

vsevel:
it is failing on the integration test module:

`Failed to execute goal io.quarkus:quarkus-maven-plugin:3.34.3:go-offline (default-cli) on project quarkus-integration-test-artemis-jms-ra: ...`

`--also-make-dependents` causes Maven to run this goal on integration-tests AND all its child modules. Since quarkus-integration-test-artemis-jms-ra is  a child of integration-tests, it's included.

vsevel:
if I execute `mvn  --batch-mode   --define maven.buildNumber.skip de.qaware.maven:go-offline-maven-plugin:resolve-dependencies` after a `mvn clean` it fails.
But after a `mvn package` it succeeds.
`quarkus:go-offline` was added back then in https://github.com/quarkiverse/quarkus-artemis/commit/2e123cce2897a666fd96bcebbb9ad5965f452b3f
do we even need this goal? (I am not sure why, it was never added to the vault or freemarker extensions. why do we need it on artemis?

holly-cummins:
> `--also-make-dependents` causes Maven to run this goal on integration-tests AND all its child modules. Since quarkus-integration-test-artemis-jms-ra is a child of integration-tests, it's included.

Ahhh, that makes sense. Thank you for spending the time to actually think about it properly :) 

So yes, I think the two options are your suggestion of dropping the make dependents and doing it in two steps, or just dropping `go-offline`. I have no idea why Artemis might be running that goal, I assumed you knew! If no one knows dropping it might be simplest, but I suppose a quick run through the history to see if the reason was written down might be sensible. 

I don't think many projects run `go-offline`, so Artemis is providing valuable coverage/acting as a useful sacrificial guinea pig here. But it's also making stuff unnecessarily hard.

turing85:
> I have no idea why Artemis might be running that goal, I assumed you knew!

This one I can answer (since I added it): the idea is that we fetch all dependencies before the matrix-build (we build each integration-test module in parallel with `--also-make` to speed up builds; otherwise a singular build would take over 1 hour). Fetching all dependencies beforehand means we do not need to download anything during the matrix-build.

holly-cummins:
Thanks, @turing85! So it sounds like removing the `-also-make-dependencies` and doing package first is the way to go?

turing85:
> Thanks, @turing85! So it sounds like removing the `-also-make-dependencies` and doing package first is the way to go?

Not sure. Since we have a lot of modules, this might take too long. Let me play around with it a little. We may drop `quarkus:go-offline` all together.

turing85:
@vsevel can you please apply the following patch to your branch? This should fix the issue by removing `quarkus:go-offline` and we can merge this PR.

```diff
Subject: [PATCH] Remove quarkus:go-offline
---
Index: .github/workflows/build.yml
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
--- a/.github/workflows/build.yml	(revision 8a45bfba369a037d7ca53abc34a47c418e13f90b)
+++ b/.github/workflows/build.yml	(revision fcecb09e0ded00b71058c08d7423ac7200c5a3fc)
@@ -110,12 +110,6 @@
             --batch-mode \
             --define maven.buildNumber.skip \
             de.qaware.maven:go-offline-maven-plugin:resolve-dependencies
-          ./mvnw \
-            --batch-mode \
-            --define maven.buildNumber.skip \
-            --projects integration-tests \
-            --also-make-dependents \
-            quarkus:go-offline
 
   build-and-test-jvm:
     runs-on: ubuntu-latest
```

turing85 (approved):

turing85:
Thanks @vsevel for your patience.

holly-cummins:
Hurray! Thanks both for figuring it out.

vsevel:
FYI I tested the new ra dev service with one of our application, and it is all working as expected, with or without compose.

holly-cummins:
> FYI I tested the new ra dev service with one of our application, and it is all working as expected, with or without compose.

Yaaaaay!

turing85:

PR #1107, opened 2026-04-16
Migrate Core Dev Service to the new API

### LINKED ISSUE #874: Replace deprecated `RunningDevService`

vsevel:
Migrate according to https://quarkus.io/blog/new-dev-services-api/, taking the RA Dev Service as an example.
Migrated using claude. I reviewed the code, and compared to the new RA dev service. everything looks correct. all tests succeed.
however I did not (and will not) try on a separate custom dedicated application. and I did not test with compose, as it is not my use case. if anybody wants to step in and make additional tests, that would be great.


docs/modules/ROOT/pages/includes/attributes.adoc
```diff
@@ -1,4 +1,4 @@
-:quarkus-version: 3.34.3
+:quarkus-version: 3.34.4
```
zhfeng: @vsevel Please ignore this change. I think we only change it when doing a new release?

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -108,116 +99,89 @@ List<DevServicesResultBuildItem> startArtemisDevService(
...
-            };
-            closeBuildItem.addCloseTask(closeTask, true);
-        }
-        cfgs.put(name, Objects.requireNonNull(configuration));
 
-        if (devServices.get(name).isOwner()) {
-            LOGGER.infof("Dev Services for ActiveMQ Artemis and named configuration %s started on %s",
-                    name, getArtemisUrl(name));
+            boolean useSharedNetwork = DevServicesSharedNetworkBuildItem.isSharedNetworkRequired(devServicesConfig,
+                    devServicesSharedNetworkBuildItem);
```
zhfeng, line 125: The old code produced `DevservicesCardBuildItem` with custom DevUI card links (Artemis Web UI with binoculars icon, broker URL with plug icon) via the deleted `createLinksToArtemis()` method. With this migration, `DevservicesCardBuildItem` is no longer produced from `startArtemisDevService`, and the `createLinksToArtemis()` method is removed.

The `createLinksToArtemis` build step (line 56-63) still exists and consumes `Optional<DevservicesCardBuildItem>`, but since nothing produces it anymore, the Optional will always be empty — unless the new Dev Services API auto-generates it from `DevServicesResultBuildItem`.

Could you verify that the DevUI links to the Artemis Management Console still work after this migration? If the new framework does not auto-generate the cards, users will silently lose the Artemis Web UI links in DevUI.

zhfeng: The AI tool find this issue and @vsevel can you take a look to see if it is a real one?

vsevel: the comment is relevant. with my implementation, the links disappear. it is going to take more time ;-)
I suspect this is broken too in the ra dev service as well. I need to understand how this is supposed to work when using the new api. any help welcome.

vsevel: I pushed a change, which does not work. the card build item gets created before the container gets started. and as such the urls are missing.
@holly-cummins what is the pattern to handle that kind of use case?

vsevel: cc @ozangunalp

holly-cummins: I hit this a few months ago, so it's documented at https://quarkus.io/blog/new-dev-services-api/#how-can-i-publish-links-to-my-service-in-the-dev-ui
Basically, it's pretty easy, you just need to know the (non-obvious) pattern.

(It's also in a PR for the 'main' guide, but that's still in review.)

vsevel: I missed that @holly-cummins . will try. thanks

vsevel: I have gone open step further. 
@turing85 on the existing dev service I noticed that when you put your mouse over `<default> web UI`, or `<default> broker URL`, you get the same url, which is the console url . in my case `http://localhost:54243/`. is this expected that? or is this an oversight of the original code:
```
    private static List<PageBuilder<?>> createLinksToArtemis() {
        List<PageBuilder<?>> result = new ArrayList<>();
        for (var name : devServices.keySet()) {
            String artemisWebUiUrl = getArtemisWebUiUrl(name);
            result.add(Page.externalPageBuilder(name + " web UI")
                    .url(artemisWebUiUrl, artemisWebUiUrl)
                    .doNotEmbed()
                    .isHtmlContent()
                    .icon("font-awesome-solid:binoculars"));
            result.add(Page.externalPageBuilder(name + " broker URL")
                    .url(artemisWebUiUrl)
                    .doNotEmbed()
                    .isHtmlContent()
                    .staticLabel(getArtemisUrl(name))
                    .icon("font-awesome-solid:plug"));
        }
        return Collections.unmodifiableList(result);
    }
```

`artemisWebUiUrl`  appears in both pages. was it the intent?

now back to the new api. I cannot get the `web-ui-url` , because there is no such property in the running app. specifically in the `ArtemisDevUIJsonRpcService` when I ask for `quarkus.artemis.web-ui-url` I get null.
**should I remove the `web-ui-url` card? if not how do we solve the fact that there are is `web-ui-url` property?**

as far as the broker url is concerned, the card appears slightly differently:
- the mouse over `<default> broker URL` shows `tcp://localhost:54244` (which is the value of config property `quarkus.artemis.url` ), and that seems correct
- the blue label on the right shows `quarkus.artemis.url` , and not `tcp://localhost:54244` like it does today. 

**are we okay with those changes? if not how can we do better?**

lastly, is this really exposing an additional non secured endpoint to retrieve config information just for the the sake of dev mode? if that ok from a security standpoint? if so, we should probably limit the scope of what can be searched. you thoughts on this @holly-cummins?

holly-cummins: That's an interesting point, @vsevel. The endpoint should only be present in dev mode (I just had a panic I forgot the guards in my example, but I didn't, luckily). 

```
 @BuildStep(onlyIf = IsLocalDevelopment.class)
    public JsonRPCProvidersBuildItem createJsonRPCService() {
```

You're right that limiting open endpoints is ideal, but in dev mode, we're much more permissive. The config information would already be available to anyone browsing `q/dev-ui`, so this is just an extra source of the same information. I think, in general, security is not the intention of dev mode. :) 

I'm not sure I understand what you mean about the scope of the information; the grpc endpoint added for this would only return the dev service url, and my assumption is that there wouldn't be any other endpoints unless they're added explicitly. 

@phillip-kruger, have I missed anything about how it works?

vsevel: claude's code suggestion is (which is the version I pushed so far):
```
@ApplicationScoped
public class ArtemisDevUIJsonRpcService {

    public String getUrl(String configKey) {
        return ConfigProvider.getConfig().getOptionalValue(configKey, String.class).orElse(null);
    }
}
```
if this was exposed in the prod application, then anybody could ask any property, including sensitive properties. as I understand it, this endpoint is not exposed outside the scope of dev mode, so we are fine.

holly-cummins: Ah, no, that's not quite right, I don't think. As you have it, it *would* be exposed. 

The `@ApplicationScoped` isn't supposed to be there, because it's the build step which causes the method to be registered.

holly-cummins: Have a look at https://github.com/quarkusio/quarkus/blob/main/extensions/oidc/runtime-dev/src/main/java/io/quarkus/oidc/runtime/dev/ui/OidcDevJsonRpcService.java, which registers the URL for keycloak. It's not registered with bean annotations, just in the build processor in a build step:

```
    @BuildStep(onlyIf = IsLocalDevelopment.class)
    JsonRPCProvidersBuildItem produceOidcDevJsonRpcService() {
        return new JsonRPCProvidersBuildItem(OidcDevJsonRpcService.class);
    }
```

vsevel: funny thing is I did not have it initially. then I saw in core:
```
@ApplicationScoped
public class CacheJsonRPCService {
```

so I added the annotation. now removing it...

turing85: > I have gone open step further. @turing85 on the existing dev service I noticed that when you put your mouse over `<default> web UI`, or `<default> broker URL`, you get the same url, which is the console url . in my case `http://localhost:54243/`. is this expected that? or is this an oversight of the original code:
> 
> ```
>     private static List<PageBuilder<?>> createLinksToArtemis() {
>         List<PageBuilder<?>> result = new ArrayList<>();
>         for (var name : devServices.keySet()) {
>             String artemisWebUiUrl = getArtemisWebUiUrl(name);
>             result.add(Page.externalPageBuilder(name + " web UI")
>                     .url(artemisWebUiUrl, artemisWebUiUrl)
>                     .doNotEmbed()
>                     .isHtmlContent()
>                     .icon("font-awesome-solid:binoculars"));
>             result.add(Page.externalPageBuilder(name + " broker URL")
>                     .url(artemisWebUiUrl)
>                     .doNotEmbed()
>                     .isHtmlContent()
>                     .staticLabel(getArtemisUrl(name))
>                     .icon("font-awesome-solid:plug"));
>         }
>         return Collections.unmodifiableList(result);
>     }
> ```
> 
> `artemisWebUiUrl` appears in both pages. was it the intent?
> 
> ...

Just an oversight. Feel free to fix this if you want.

vsevel: the most pressing concern is not being able to get the web ui url at this point.
I tried adding `Optional<String> webUiUrl()` in the runtime config, but in spite of 
``` 
                        .configProvider(Map.of(
                                urlPropertyName,
                                container -> String.format("tcp://%s:%d", container.getHost(), container.getPort()),
                                webUiUrlPropertyName,
                                container -> String.format("http://%s:%d", container.getHost(), container.getWebUiPort())))
``` 
I still can't get `quarkus.artemis.web-ui-url` being correctly valued in the app.

as a result, this does can't find the web ui url:
``` 
            CardPageBuildItem cardPage = new CardPageBuildItem();
            cardPage.addPage(Page.externalPageBuilder(name + " web UI")
                    .dynamicUrlJsonRPCMethodName("getUrl", Map.of("configKey", webUiUrlPropertyName))
                    .doNotEmbed()
                    .isHtmlContent()
                    .icon("font-awesome-solid:binoculars"));
```

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -108,116 +98,102 @@ List<DevServicesResultBuildItem> startArtemisDevService(
...
+
+            DevServicesResultBuildItem discovered = discoverRunningService(composeProjectBuildItem, name, configuration,
+                    urlPropertyName, launchMode.getLaunchMode(), useSharedNetwork);
+            if (discovered != null) {
+                devServicesResult.produce(discovered);
+            } else {
+                Optional<Duration> timeout = devServicesConfig.timeout();
+                devServicesResult.produce(DevServicesResultBuildItem.<ArtemisContainer> owned()
+                        .name("ActiveMQ-Artemis " + name)
```
turing85: This method is deprecated. According to its documentation, `feature(String)` should be used instead.

core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
```diff
@@ -108,116 +98,102 @@ List<DevServicesResultBuildItem> startArtemisDevService(
...
+            ArtemisDevServiceCfg config,
+            String urlPropertyName,
+            LaunchMode launchMode,
+            boolean useSharedNetwork) {
+        return artemisContainerLocator.locateContainer(config.serviceName, config.shared, launchMode)
+                .or(() -> ComposeLocator.locateContainer(composeProjectBuildItem,
+                        List.of(config.imageName, "artemis"),
+                        ARTEMIS_PORT, launchMode, useSharedNetwork))
+                .map(containerAddress -> DevServicesResultBuildItem.discovered()
+                        .name("ActiveMQ-Artemis " + name)
```
turing85: This method is deprecated. According to its documentation, `feature(String)` should be used instead.

turing85 (changes requested):
I think the same change is necessary in the RA test resource.

phillip-kruger:
W.r.t The JsonRPCService:

- Default scope is `ApplicationScope`, so except if you want to change the scope there is no need for the annotation (although it does no harm)
- The processor will only make this a CDI bean in Dev mode due to the `@BuildStep(onlyIf = IsLocalDevelopment.class)`
- You can also add this to the `runtime-dev` module (vs `runtime`) that will mean that the class will not be in the final prod artefact. See https://quarkus.io/guides/dev-ui#jsonrpc-against-the-runtime-classpath, especially the part at the end of that section explaining the pom changes needed in `runtime`.

W.r.t The URL:
You might not need a JsonRPC Service at all.
```
cardPage.addPage(Page.externalPageBuilder(name + " web UI")                                                                                                                 
      .dynamicUrlJsonRPCMethodName("devui-dev-services:devServicesConfig",                                                                                                    
          Map.of("name", "artemis",       // must match .feature("artemis") on the builder                                                                                    
                 "configKey", webUiUrlPropertyName))                                                                                                                          
      .doNotEmbed()                                                                                                                                                           
      .isHtmlContent()                                                                                                                                                        
      .icon("font-awesome-solid:binoculars"));
```

The feature name used in DevServicesResultBuildItem.owned().feature(...), and the configKey must match exactly what's in .configProvider(Map.of(webUiUrlPropertyName, ...))

turing85 (commented):
The compiler still complains about `ra`: 

```
Warning:  /home/runner/work/quarkus-artemis/quarkus-artemis/ra/deployment/src/main/java/io/quarkus/artemis/jms/ra/deployment/DevServicesArtemisProcessor.java:[101,21] name(java.lang.String) in io.quarkus.deployment.builditem.DevServicesResultBuildItem.OwnedServiceBuilder has been deprecated
Warning:  /home/runner/work/quarkus-artemis/quarkus-artemis/ra/deployment/src/main/java/io/quarkus/artemis/jms/ra/deployment/DevServicesArtemisProcessor.java:[136,25] name(java.lang.String) in io.quarkus.deployment.builditem.DevServicesResultBuildItem.DiscoveredServiceBuilder has been deprecated
```

Can/Should we fix this in this PR?

vsevel:
@turing85 I resolved both deprecated method calls. once this PR gets merged, I will take a look at the RA version.
@phillip-kruger a big thanks for your help. this is much simpler now
I tested with an app that contains a default CF, plus a named one. they both appear as expected on the dev ui for extension Artemis Core.
With this PR, the "Artemis JMS" extension does not show the links anymore, because they used to be produced by consuming the `DevservicesCardBuildItem`, which are not produced anymore. as a result the links are shown only on the "Artemis Core" extension.

turing85 (approved):

vsevel:
I am already working on the follow up for the RA.
I am waiting for that one to get merged.
let me know if you need anything else.

vsevel:
> The compiler still complains about `ra`:
> 
> ```
> Warning:  /home/runner/work/quarkus-artemis/quarkus-artemis/ra/deployment/src/main/java/io/quarkus/artemis/jms/ra/deployment/DevServicesArtemisProcessor.java:[101,21] name(java.lang.String) in io.quarkus.deployment.builditem.DevServicesResultBuildItem.OwnedServiceBuilder has been deprecated
> Warning:  /home/runner/work/quarkus-artemis/quarkus-artemis/ra/deployment/src/main/java/io/quarkus/artemis/jms/ra/deployment/DevServicesArtemisProcessor.java:[136,25] name(java.lang.String) in io.quarkus.deployment.builditem.DevServicesResultBuildItem.DiscoveredServiceBuilder has been deprecated
> ```
> 
> Can/Should we fix this in this PR?

this will be in the follow up PR, along with links in dev ui

zhfeng (approved):
Thanks @vsevel - LGTM

holly-cummins:
> W.r.t The URL: You might not need a JsonRPC Service at all.
> 
> ```
> cardPage.addPage(Page.externalPageBuilder(name + " web UI")                                                                                                                 
>       .dynamicUrlJsonRPCMethodName("devui-dev-services:devServicesConfig",                                                                                                    
>           Map.of("name", "artemis",       // must match .feature("artemis") on the builder                                                                                    
>                  "configKey", webUiUrlPropertyName))                                                                                                                          
>       .doNotEmbed()                                                                                                                                                           
>       .isHtmlContent()                                                                                                                                                        
>       .icon("font-awesome-solid:binoculars"));
> ```

Ohh, I didn't know this! I should add it to the docs. 
> 
> The feature name used in DevServicesResultBuildItem.owned().feature(...), and the configKey must match exactly what's in .configProvider(Map.of(webUiUrlPropertyName, ...))

vsevel:
@holly-cummins are you ok with the proposed changes?

holly-cummins (approved):

turing85:

PR #1110, opened 2026-04-20
Fix NPE in ArtemisResourceAdapterFactory when connection-parameters is not set

gastaldi:
The count() method did not handle a null text argument, causing a `NullPointerException` when the `connection-parameters` config key is absent. Return 0 for `null` input so a single connector is used by default.


zhfeng (approved):

turing85:
@gastaldi  this change has been released with version `3.14.2` ans is now [available on central (`central.sonatype.com`)](https://central.sonatype.com/artifact/io.quarkiverse.artemis/quarkus-artemis-parent/3.14.2)

PR #1111, opened 2026-04-21
Dev Service Artemis JMS Resource Adapter links

vsevel:
this PR adds the links to the RA dev service, and corrects the 2 calls to deprecated `name(String)`  method.
given the layout, and the size of the property names (e.g. `quarkus.ironjacamar.ra.config.connection-parameters`), this looks a bit ugly. the property name has some value, so I am not sure whether or not we want to keep it. your opinion?
the url for the console ui works fine. but the property value for the broker connection parameters `host=localhost;port=56482;protocols=CORE` is rendered as
`http://localhost:8080/q/dev-ui/host=localhost;port=56482;protocols=CORE`, which is wrong.
I think this is caused by the connection parameters not having an `http` scheme (well, it has no scheme altogether).
I am not sure we can do better. any idea @phillip-kruger ?
other than that, the change works fine. I tested it on an app with 2 ra:
```
quarkus.ironjacamar.ra.kind=artemis
quarkus.ironjacamar.foo.ra.kind=artemis
```
```
        <dependency>
            <groupId>io.quarkiverse.artemis</groupId>
            <artifactId>quarkus-artemis-jms-ra</artifactId>
        </dependency>
```
I also corrected typos in the documentation.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/artemis/core/deployment/DevServicesArtemisProcessor.java
@@ -47,7 +47,7 @@ public class DevServicesArtemisProcessor {
      */
     static final String DEV_SERVICE_LABEL = "quarkus-dev-service-artemis";
     static final int ARTEMIS_PORT = 61616;
-    static final int ARTEMIS_WEB_UI_PORT = 8161;
+    public static final int ARTEMIS_WEB_UI_PORT = 8161;
 
     private static final ContainerLocator artemisContainerLocator = locateContainerWithLabels(ARTEMIS_PORT,
             DEV_SERVICE_LABEL);
@@ -196,7 +196,7 @@ public class DevServicesArtemisProcessor {
         return getArtemisPropertyBase(name) + "url";
     }
 
-    private static String getWebUiUrlPropertyName(String name) {
+    public static String getWebUiUrlPropertyName(String name) {
         return getArtemisPropertyBase(name) + "web-ui-url";
     }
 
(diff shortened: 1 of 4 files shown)


phillip-kruger:
What about using the static label to show the details ?

```
cardPage.addPage(Page.externalPageBuilder(name + " web UI")                                                                                                                 
          .dynamicUrlJsonRPCMethodName("devui-dev-services:devServicesConfig",                                                                                                
                  Map.of("name", feature, "configKey", webUiUrlPropertyName))                                                                                                 
          .doNotEmbed()                                                                                                                                                       
          .isHtmlContent()                                                                                                                                                    
          .staticLabel(urlPropertyName)   // shows the config key name                                                                                                        
          .icon("font-awesome-solid:binoculars"));
```

vsevel:
what details exactly? I probably misunderstood what you were suggesting.
this is what I have right now:
image
and I move the mouse over to the broker URL :
image
the first issue is long parameter name. I do not see any solution other thn removing it, or accepting it leaks on the card on the right
the second issue is when we mouse over the broker url value, then we have an incorrect http url, insteed of just the connection parameteres. not sure if we can do better.

vsevel:
dear all, I found that we were missing the extra arg `--relax-jolokia` . this has been fixed.
I have also removed the second card `broker URL`. as explained before this config parameter did not fit well within the current layout. and the lack of scheme would make it appear as a path relative to the dev-ui url, which is wrong.
now we have one label per resource adapter.
I think this is ready for review @turing85

phillip-kruger:
Yeah w.r.t the dev ui card, maybe the only option for now is a custom card

turing85 (approved):

turing85:
@vsevel PR is approved. Do you want to change something, or can we merge?

vsevel:
We can go ahead. Thanks @turing85

turing85:

PR #1205, opened 2026-08-18
General housekeeping

turing85:
- use records where applicable
- remove unnecessary casts
- in streams, use ".toList()" instead of ".collect(Collectors.toList())"
- Remove explicit type parameters where the type can be inferred
- use lambdas instead of anonymous interface implementations where possible
- Use method references instead of lambdas where applicable

### DIFF
--- a/jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/tracing/TracingMessageConsumer.java
+++ b/jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/tracing/TracingMessageConsumer.java
@@ -34,7 +34,7 @@ class TracingMessageConsumer implements MessageConsumer {
 
     @Override
     public Message receive() throws JMSException {
-        return receiveWithTracing(() -> delegate.receive());
+        return receiveWithTracing(delegate::receive);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


turing85:
@middagj @zhfeng I'd like a review from you on this PR.

The motivation for this PR was: I ran IntelliJ's code inspection and fixed what it found.

zhfeng (approved):
LGTM

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/tracing/TracingMessageConsumer.java
```diff
@@ -119,19 +119,10 @@ public void close() throws JMSException {
...
-        private final JmsContextPropagator contextPropagator;
-
-        TracingMessageListener(MessageListener delegate, Destination destination, Tracer tracer,
-                JmsContextPropagator contextPropagator) {
-            this.delegate = delegate;
-            this.destination = destination;
-            this.tracer = tracer;
-            this.contextPropagator = contextPropagator;
-        }
+    private record TracingMessageListener(MessageListener delegate, Destination destination,
```
middagj, line 122: Should this really be a record? I thought record where intended for data like objects, not just for an object that have only final variables.

middagj: Ah, we already seem to do it like that in other cases

middagj (approved):
Small question out of curiosity

jms/runtime/src/main/java/io/quarkus/artemis/jms/runtime/tracing/TracingMessageConsumer.java
turing85: We didn't use records before this PR. I introduced them.
For me, if all fields should be unmodifiable (and the system allows it), then records should be used (shorter synta for what is wanted).

